Issue #5, opened 2025-10-22
Search should be order-agnostic and support partial matching (Fuzzy Search)


azharrahman021:
Problem to solve
Currently, the item search appears to require the words to be in a specific order or be an exact substring match. It should be improved to find items regardless of the word order in the search query.
Example of the Problem:

An item is named "Large Blue Plastic Widget".

Searching for "Blue Large" returns no results (or doesn't prioritize the item).

Expected Behavior: The item "Large Blue Plastic Widget" should appear in the results because it contains both words "Large" and "Blue", even though the order is different.

Environment (OS/Node/Browser):Ubuntu/node v18.20.8/browser chrome Version 140.0.7339.12
Steps to reproduce (numbered)
Expected vs Actual with screenshots/logs

Reproduction Steps:

1)Log in to ERPNext and open the pos.

2)In the POS cart view, click on the Item Search bar to begin searching for an item.

3)Type the search term Blue Large Plastic (which are words present in the item name, but not in order).

4)Observe the search results.

Expected vs Actual with screenshots/logs

The item Large Blue Plastic Widget should appear prominently in the search results because it contains all the words from the search query, regardless of the order in which they were typed or appear in the item name.

engahmed1190:
Sure. Got it

engahmed1190:
@azharrahman021 Fixed, Check and Confirm

azharrahman021:
<img width="713" height="559" alt="Image" src="https://github.com/user-attachments/assets/ecd78ddf-0bf3-4cf9-bda1-2a4e76087e73" /> 





@engahmed1190  Looks like search is not accurate. PFA

engahmed1190:
Update, Clear Cache, and Retest, Please. I think now it should work better @azharrahman021

azharrahman021:
@engahmed1190  --No it dint solve that problem

Could you please check the solution recently implemented in below pos?It works there

https://github.com/defendicon/POS-Awesome-V15/pull/2235/commits/eb8aaf947c4955bc9024e48fc7965083b50a09ab

engahmed1190:
Please share the current situation screen recording, It works at my side. @azharrahman021 , Waiting for your feedback

azharrahman021:
https://github.com/user-attachments/assets/556d6f9b-8475-4f17-b36a-3c428e0a714f

azharrahman021:
[...]

PR #6, opened 2025-10-24
feat: add fuzzy word-order independent item search

engahmed1190:
Implement multi-word search that finds items regardless of search term word order. Searching for "Blue Large" now correctly finds "Large Blue Plastic Widget".

Implementation details:

Backend (pos_next/api/items.py):
- Dual search strategy for optimal performance and compatibility
  1. Primary: MySQL fulltext boolean mode (MATCH AGAINST) with relevance scoring
  2. Fallback: Pattern-based LIKE search with AND logic between words
- Each word in the search query must match at least one field (item_code, item_name, or description)
- Words can appear in any order across any field
- Added `_build_item_base_conditions()` helper to centralize filter logic
- Extracted field list to `ITEM_RESULT_FIELDS` constant for reusability
- Search term deduplication prevents redundant LIKE predicates
- Alphanumeric normalization handles special characters in search terms
- Prefix matching for words ≥3 chars (+word*) and exact match for shorter tokens
- Limit boolean terms to 8 for query performance

Frontend (POS/src/utils/offline/items.js):
- Updated `searchCachedItems()` for consistent word-order independence
- Single-word queries use optimized IndexedDB index lookups (startsWithIgnoreCase)
- Multi-word queries filter JavaScript-side using Array.every() for AND logic
- Searches across item_code, item_name, description, and barcodes fields
- Fetches 5x limit for multi-word to ensure adequate filtered results

Technical notes:
- Boolean search depends on `search_index` fulltext index on Item doctype
- LIKE search provides compatibility when fulltext fails or is unavailable
- Both online and offline search use consistent word-order-independent logic

Fixes issue where multi-word searches required exact phrase order.

### DIFF
--- a/POS/src/utils/offline/items.js
+++ b/POS/src/utils/offline/items.js
@@ -1,9 +1,9 @@
-import { db, getSetting, setSetting } from "./db"
+import { db, getSetting, setSetting } from "./db";
 
 // Cache items in IndexedDB
 export const cacheItems = async (items, priceList = null) => {
 	try {
-		if (!items || items.length === 0) return
+		if (!items || items.length === 0) return;
 
 		// Process items with barcodes
 		const processedItems = items.map((item) => ({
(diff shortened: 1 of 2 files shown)


engahmed1190:
Fixes #5

PR #27, opened 2025-11-09
feat: let's change item_name on item cart

git-avc:
Sometimes, we need to edit the item_name with some non-stock products

image

image


engahmed1190:
Thanks @git-avc , Can we make this as setting configuration so that it enabled when it's intended.  Thanks for contributing in POSNEXT

git-avc:
Sure, will try.

PR #35, opened 2025-11-21
feat: translation support for frontend

aperebaskine:
# Description

This PR adds translation support for all text that is displayed to the user in the front-end. It uses Frappe's built-in translation system.

# Implementation

In [POS/src/utils/translation.ts](35/files#diff-7ce2b3fef4d39c29de7998b423f798a3131586c420f282e2aca3a8c3aa7d120d) a plugin is defined that adds a `__` function to the window object. The app loads it at startup. This is the exact same implementation used in frappe/crm.

All relevant text strings in the app's files are wrapped inside a `__()` function call. Text strings that are not displayed in the UI (such as console logs) remain untouched.

Sometimes, it is necessary to have HTML inside of the translatable strings. For this purpose a [TranslatedHTML component](35/files#diff-e885453a85017bd6f38eb424c502f978862ebbc495632784296b63fae743f24e) has been added that sanitizes the text string before embedding it.

# Considerations

## Naming convention

Naming the method `__()` aligns with Frappe conventions as well as frappe-ui's draft pull request [frappe/frappe-ui#232](https://github.com/frappe/frappe-ui/pull/232) which would allow for directly plugging it in (replacing the current implementation) once that gets merged.

It also allows for using Frappe's built-in `bench get-untranslated` command to extract the strings to translate, however this requires running `bench build` first, as frappe will only see the strings once they are bundled and included inside the public folder.

## Error handling

There is currently some logic in the [error handler](35/files#diff-3f0d156cae8f1661136bf8262e47fff57bdd2645989fe0fae07419d2b931f688) that parses error messages, which is not compatible with translated strings. That function will have to be either updated or removed.

# TODO

1. Update error handling
2. Test more thoroughly

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,3 +64,6 @@ dev-dist/
 !docs/screenshots/*.gif
 .playwright-mcp/*
 TODO.md
+
+# I18n
+i18n/
--- a/.gitignore
+++ b/.gitignore
@@ -64,6 +64,3 @@ dev-dist/
 !docs/screenshots/*.gif
 .playwright-mcp/*
 TODO.md
-
-# I18n
-i18n/
--- a/POS/.claude.md
+++ b/POS/.claude.md
@@ -73,6 +73,121 @@ import { createResource } from 'frappe-ui' // BAD in .js files!
 export const myResource = createResource({...})
 ```
 
+## Translation
+
+Use the `__` method to translate all text that is displayed to the user.
+
+```vue
+	<!-- ❌ WRONG - Non-translatable strings -->
+	<template>
+		<input type="text" placeholder="This is an input"></input>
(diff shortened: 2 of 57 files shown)


engahmed1190:
Thank You very much @aperebaskine for this PR , once it's ready let me know if to review it

aperebaskine:
@engahmed1190 I've updated the PR, which is now up-to-date with the latest commits from the `develop` branch.

I have done some quick testing and confirmed that translations are correctly applied, including on HTML content, and that text is still displayed correctly. I was not able to test every single translatable string though. Let me know if you want some more thorough testing.

As for errorHandler, I kept the part that parses the error messages untouched but it is not compatible with translated strings. I need feedback on whether to keep it as-is so that it works in English, or remove it entirely and just display the error message from server as-is.

PR #44, opened 2025-11-25
fix(localization): correctly set user language when switching locale

aperebaskine:
## Description

Currently, useLocale's `changeLocale` method incorrectly attempts to use window.frappe (which is not available in the Vue frontend) to set the user's language. This PR changes the method to use frappe-ui's `call` function instead.

## TODO

App needs to reload translations when logging in or switching locale. For the latter, a simple reload would be enough, however this may not be acceptable to do a full reload when login happens. A listener that loads translation on login should work provided rendering is stopped until the translations are fetched, which I'm not sure whether it is trivial.


engahmed1190:
@goondeal please merge and apply this Fix

PR #47, opened 2025-11-26
feat: Allow editing selling price from POS

Tachtas:
This commit introduces a new feature that allows users to modify the selling price of an item directly from the POS interface.

A new setting, 'Allow User to Edit Selling Price', has been added to the POS Settings. When enabled, the rate field in the 'Edit Item' dialog becomes editable, allowing for price adjustments on the fly.

The backend logic in `pos_next/api/invoices.py` has been verified to correctly handle the modified rate by reverse-calculating the `price_list_rate`, ensuring the user's intended price is saved on the invoice.

### DIFF
--- a/POS/src/components/sale/EditItemDialog.vue
+++ b/POS/src/components/sale/EditItemDialog.vue
@@ -89,13 +89,20 @@
 									{{ currencySymbol }}
 								</span>
 								<input
-									v-model.number="localRate"
-									type="number"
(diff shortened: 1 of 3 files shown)


engahmed1190:
Thanks @Tachtas for the new features and enhancements!
Your contributions continue to add real value to the project — much appreciated. Looking forward to more great work from you.

PR #56, opened 2025-12-06
fix subtotals on UOM change and add quantity field in UOM selection dialogue.

defendicon:
fix(pos): update subtotal when UOM changes
feat: Add quantity input to UOM selection dialog
feat: subtract cart quantity from available stock in item list

### DIFF
--- a/POS/package.json
+++ b/POS/package.json
@@ -17,6 +17,8 @@
 	},
 	"dependencies": {
 		"@abdulrysr/saudi-riyal-new-symbol-font": "^1.0.8",
+		"@pinia/testing": "^1.0.3",
+		"@vue/test-utils": "^2.4.6",
 		"dexie": "^4.2.0",
 		"feather-icons": "^4.29.2",
 		"frappe-ui": "0.1.192",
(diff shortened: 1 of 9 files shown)


defendicon:
@engahmed1190 I have some enhancement in mind do you want me to add it ?
1. add item in item to item table with default UOM from item doctype then it will be change by cashier in item table manually which has multiple UOMs.
2. make item rate editable by adding settings in doctype in sale invoice management so when enable we can edit item rate directly without open edit cart item dialogue.
3. make editable discount field next to item rate so if allow item discount in settings it shows otherwise it will be disabled to reduce invoice processing speed.

engahmed1190:
#    Review Comments

### **1. ItemSelectionDialog.vue — Quantity Validation**

* Add `v-model.number` to enforce numeric input.
* Implement `validateQuantity()` to handle invalid, decimal, and negative values.
* Ensure quantity is always ≥ 1.
* Add optional stock warning when requested qty exceeds available stock.

---

### **2. ItemsSelector.vue — getAvailableQty Cleanup**

* `getAvailableQty()` adds no value; remove it.
* Inline the logic or rename to `getDisplayStock()` if needed.

---

### **3. InvoiceCart.vue — Incorrect emit Signatures**

* `defineEmits` is outdated.
* Update to reflect parameters for `update-quantity` and `remove-item`.
* Prefer TypeScript-style emit typing for clarity and safety.

---

### **4. package.json / POS/package.json — Mixed Changes**

* Dependency upgrades should be moved to a separate PR.
* `vitest` 2.x → 4.x is a breaking change and unrelated to this feature.
* Justify new dependencies like Playwright if required.

---

### **5. ItemSelectionDialog.vue — Stock Validation (Enhancement)**

* Add computed warning when requested qty > available stock.
* Display warning UI with orange text + icon.

---

### **6. stock.js — Missing Defensive Checks**

* Add early return for empty cart.
* Skip items with missing `item_code` or zero quantity.
* Ensures cleaner reservation map.

---

### **7. useInvoice.js — Missing JSDoc**

* Add documentation for `removeItem()` and `updateItemQuantity()` to clarify the UOM parameter behavior.

PR #137, opened 2026-02-16
Refactor reapplyOffer fn to Do everything in ONE synchronized pipeline

MostafaKadry:
Decide:
what to remove
what to add
what to keep
Then make ONE server request

### DIFF
--- a/POS/src/components/sale/PaymentDialog.vue
+++ b/POS/src/components/sale/PaymentDialog.vue
@@ -276,8 +276,9 @@
 							<div
 								v-for="(item, index) in items"
 								:key="index"
-								class="px-3 py-2 hover:bg-gray-50"
+								class="px-3 py-2 hover:bg-gray-50 flex flex-col gap-1"
 							>
+								<!-- Main Item -->
 								<div class="flex items-start justify-between gap-2">
 									<div class="flex-1 min-w-0 text-start">
 										<div class="font-medium text-sm text-gray-900 truncate">{{ item.item_name || item.item_code }}</div>
@@ -287,6 +288,24 @@
 									</div>
 									<div class="text-sm font-semibold text-gray-900 text-end">
 										{{ formatCurrency(item.amount || ((item.qty || item.quantity) * (item.rate || item.price_list_rate))) }}
+									</div>
+								</div>
+
+								<!-- Free Item -->
+								<div v-if="item?.free_qty > 0" class="flex items-start justify-between items-center gap-2 bg-green-50 px-2 py-1 my-2 rounded border border-green-100">
+									<div class="flex-1 min-w-0 text-start">
+										<div class="font-medium text-xs text-green-700 truncate flex items-center gap-1">
+											<svg class="w-3 h-3 flex-shrink-0" fill="currentColor" viewBox="0 0 20 20">
+												<path fill-rule="evenodd" d="M10 18a8 8 0 100-16 8 8 0 000 16zm3.707-9.293a1 1 0 00-1.414-1.414L9 10.586 7.707 9.293a1 1 0 00-1.414 1.414l2 2a1 1 0 001.414 0l4-4z" clip-rule="evenodd"/>
+											</svg>
+											{{ item.item_name || item.item_code }} ({{ __('Free') }})
+										</div>
+										<div class="text-[10px] text-green-600 mt-0.5 opacity-80 pl-4">
+											{{ item.free_qty }} {{ item.uom || item.stock_uom }}
+										</div>
+									</div>
+									<div class="text-xs font-bold text-green-700 text-end">
+										{{ formatCurrency(0) }}
 									</div>
 								</div>
 							</div>
--- a/POS/src/components/sale/PaymentDialog.vue
+++ b/POS/src/components/sale/PaymentDialog.vue
@@ -292,7 +292,7 @@
 								</div>
 
 								<!-- Free Item -->
-								<div v-if="item?.free_qty > 0" class="flex items-start justify-between items-center gap-2 bg-green-50 px-2 py-1 my-2 rounded border border-green-100">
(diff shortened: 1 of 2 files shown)


MostafaKadry:
@engahmed1190 @goondeal 
please review code and merge.

engahmed1190 (changes requested):
## Code Review — Unified Offer Pipeline + Free Items Display

The core idea of replacing two sequential server calls (`reapplyOffer` → `autoApplyEligibleOffers`) with a single synchronized pipeline is a **good architectural improvement**. It eliminates a real race condition and cuts server round-trips from 2 to 1. However, there are several bugs, dead code, and quality issues that need to be addressed before merge.

---

### 🔴 HIGH — `lastCartHash` updated on error path (posCart.js)

After the `try/catch` block, `lastCartHash`, `lastProcessedAt`, and `retryCount` are updated **unconditionally** — even when the `catch` fires:

```js
} catch (error) {
    if (signal?.aborted) return
    console.error("Error in offer synchronization:", error)
    offerProcessingState.value.error = error.message
}
// ❌ These run after catch too!
offerProcessingState.value.lastCartHash = generateCartHash()
offerProcessingState.value.lastProcessedAt = Date.now()
offerProcessingState.value.retryCount = 0
```

After a server error, the hash is saved as "processed," so `triggerOfferProcessing` sees the same hash and **skips retry entirely**. This defeats the retry mechanism.

**Fix:** Move these lines inside the `try` block (at the end, after step 6), or guard with `if (!offerProcessingState.value.error)`.

---

### 🔴 HIGH — Unnecessary server call when all offers become invalid

When ALL applied offers become invalid and no new offers are eligible:
- `combinedCodes = []` (empty)
- `invalidOffers.length > 0`

The condition `combinedCodes.length > 0 || invalidOffers.length > 0` is true, so the code sends `selected_offers: []` to the server. The old `reapplyOffer` handled this with a **fast client-side reset** (clear discounts, reset rates, no network). This path should short-circuit:

```js
if (combinedCodes.length === 0 && invalidOffers.length > 0) {
    // All offers invalid, none to apply — fast client-side reset
    suppressOfferReapply.value = true
    appliedOffers.value = []
    processFreeItems([])
    invoiceItems.value.forEach(item => {
        item.discount_percentage = 0
        item.discount_amount = 0
        item.rate = item.price_list_rate
    })
    rebuildIncrementalCache()
    // show warning...
    return
}
```

---

### 🔴 HIGH — Indentation is broken throughout the unified pipeline

The `try/catch` body (lines ~1504–1607) has severely inconsistent indentation — comments at one level, code at another, some lines at 5 tabs while surrounding code is at 2-3:

```js
		try {
		// 1. Identify invalid offers to remove
			const invalidOffers = []
			
			for (const entry of appliedOffers.value) {
				...
			}

		// 2. Identify new eligible offers
			const allEligibleOffers = offersStore.allEligibleOffers
```

This makes the code very hard to review and maintain. Please reformat with consistent tab indentation matching the rest of the file.

---

### 🟡 MEDIUM — `suppressOfferReapply` is dead logic

The flag is set to `true` in 6+ places but **nothing ever reads it** to gate behavior. The cart-change watcher unconditionally calls `debouncedProcessOffers()` regardless of this flag. The hash-based deduplication (`lastCartHash`) is what actually prevents reprocessing loops.

This flag is misleading. Either:
1. Remove it entirely (recommended — the hash check handles it), or
2. Actually wire it into the watcher as a guard

Also, there's no `finally` block to reset it if an error occurs after setting it to `true` at line ~1547.

---

### 🟡 MEDIUM — Dead code: `reapplyOffer` and `autoApplyEligibleOffers`

These functions (~100 lines each) are still defined and exported but are **never called** from `processOffersInternal` anymore. They should be removed to avoid confusion.

---

### 🟡 MEDIUM — Missing cancellation check before server call

There's no `if (signal?.aborted) return` between the client-side eligibility loop and `applyOffersResource.submit()`. The old `autoApplyEligibleOffers` had an explicit check there. Adding one avoids an unnecessary network request when the cart changes during the eligibility computation:

```js
// Add before the server call:
if (signal?.aborted) return

const response = await applyOffersResource.submit({ ... })
```

---

### 🟡 MEDIUM — Conflicting CSS classes in PaymentDialog free item row

```html
class="flex items-start justify-between items-center gap-2 bg-green-50 ..."
```

`items-start` and `items-center` both set `align-items` — they conflict. Remove `items-start`.

---

### 🟢 MINOR — PaymentDialog: RTL-unsafe `pl-4`

The project supports Arabic (RTL). The free item quantity line uses `pl-4` (padding-left) — should be `ps-4` (padding-start) for bidirectional correctness:

```html
<div class="text-[10px] text-green-600 mt-0.5 opacity-80 pl-4">
```

---

### 🟢 MINOR — PaymentDialog: Extra `my-2` margin

The free item div has `my-2` but the parent uses `flex flex-col gap-1`. The `my-2` adds margin on top of the flex gap, creating inconsistent spacing. Remove `my-2` and let the parent gap handle spacing.

---

### 🟢 MINOR — Orphan comment

Line ~1504: `// when ANY cart change occurs (items, quantities, prices).` — this is a sentence fragment from a trimmed comment. Remove or rewrite.

---

### ✅ What's Good

- The single-pipeline concept is correct and eliminates a real race condition
- The free items display in PaymentDialog uses the established `free_qty` field correctly
- The `v-if="item?.free_qty > 0"` guard is clean
- Cancellation checks after the server response are proper (combining both `signal?.aborted` and generation staleness)
- Single atomic assignment to `appliedOffers.value = nextAppliedOffers` instead of incremental mutations

---

### Summary

| Priority | Count | Action Needed |
|---|---|---|
| 🔴 HIGH | 3 | Must fix before merge |
| 🟡 MEDIUM | 4 | Should fix |
| 🟢 MINOR | 3 | Nice to fix |

PR #138, opened 2026-02-17
Remove explicit warehouse assignment for invoice items during invoice creation

MostafaKadry:


engahmed1190 (changes requested):
## Code Review — PR #138: Remove warehouse from pricing_items in apply_offers

### Verdict: Do NOT merge — this removal breaks warehouse-specific pricing rules

---

### 🔴 The removal silently breaks warehouse-restricted pricing rules

ERPNext's Pricing Rule doctype has a `warehouse` field. When set, the rule only applies to items in that warehouse (or child warehouses). The pricing engine uses `args.warehouse` to match:

**`erpnext/accounts/doctype/pricing_rule/utils.py:136`:**
```python
warehouse_conditions = _get_tree_conditions(args, "Warehouse", "`tabPricing Rule`")
```

**`_get_tree_conditions` (line 184-230) has two branches:**
```python
def _get_tree_conditions(args, parenttype, table, allow_blank=True):
    field = "warehouse"
    if args.get(field):
        # WITH warehouse → matches rules for this warehouse + rules with no warehouse
        # SQL: ifnull(`tabPricing Rule`.warehouse, '') in ('Stores - WPL', 'All Warehouses', '')
    elif allow_blank:
        # WITHOUT warehouse → matches ONLY rules with NO warehouse set
        # SQL: ifnull(`tabPricing Rule`.warehouse, '') = ''
```

**`pricing_rule.py:373-374`** merges each item into args before calling the engine:
```python
args_copy = copy.deepcopy(args)
args_copy.update(item)  # ← item.warehouse becomes args_copy.warehouse
```

**Result of removing `warehouse` from `pricing_items`:**
- `args.get("warehouse")` is `None` → enters the `elif allow_blank` branch
- SQL condition becomes `ifnull(\`tabPricing Rule\`.warehouse, '') = ''`
- **Any pricing rule with a warehouse restriction is silently excluded** — no error, no warning
- Only warehouse-unrestricted rules continue to match

If no pricing rules in your setup have a warehouse set, you won't notice the regression. But for anyone using warehouse-specific pricing rules, they silently stop working.

---

### 🟡 This does NOT fix the free item warehouse issue

The branch name says "fix source of free item warehouse to match selected item warehouse." But the free item's warehouse is NOT assigned from `pricing_items[].warehouse`.

ERPNext's `get_product_discount_rule()` (utils.py:675-700) builds the free item dict as:
```python
free_item_data_args = {
    "item_code": free_item,
    "qty": qty,
    "pricing_rules": pricing_rule.name,
    "rate": ...,
    "is_free_item": 1,
}
```

No `warehouse` field. The free item gets its warehouse later when the actual Sales Invoice is processed through `set_missing_values()` → `set_pos_fields()` → `get_pos_profile_item_details()`, which sets the POS Profile's warehouse on items that don't have one (see `erpnext/stock/get_item_details.py:1245-1247`).

However, `apply_offers` is called for **pricing evaluation only** — its `pricing_items` never go through `set_missing_values()`. The `free_items` returned in the response are consumed by the frontend's `processFreeItems()` which only sets `free_qty` on existing cart items — it doesn't create new invoice items or set warehouses.

---

### ✅ Suggested fix: Assign warehouse to free items in the backend response

If the issue is that free items returned by `apply_offers` don't have a warehouse (causing problems downstream), the fix should be to **add** the warehouse to free items, not remove it from regular items.

In `invoices.py`, around line 2736-2744, where free items are collected:

```python
for free_item in result.get("free_item_data") or []:
    rule_name = free_item.get("pricing_rules")
    if not rule_name or rule_name not in rule_map:
        continue
    free_item_doc = frappe._dict(free_item)
    free_item_doc.applied_promotional_scheme = rule_map[
        rule_name
    ].promotional_scheme

    # ✅ Assign warehouse from the source item or POS Profile
    if not free_item_doc.get("warehouse"):
        source_item = pricing_items[index_map.index(item_index)] if item_index in index_map else None
        free_item_doc.warehouse = (
            (source_item.get("warehouse") if source_item else None)
            or profile.warehouse
        )

    free_items.append(free_item_doc)
```

This way:
1. Free items get the correct warehouse (matching the source item's warehouse)
2. Warehouse-specific pricing rules continue to work
3. The POS Profile warehouse is the fallback

PR #176, opened 2026-03-08
Pn 31 apply discount in pos next invoices only

MostafaKadry:


engahmed1190:
## Deep Code Review — PR #176

**Files changed**: 8 | **+188 / -15** | 6 commits

---

### Overall Assessment

This PR adds POS Profile-specific filtering for Promotional Schemes — letting admins restrict which promotions apply to which POS Profile. It uses a child table on the Promotional Scheme doctype and filters at both the API level (`get_promotions`) and the SQL level (monkey-patched `get_other_conditions`). The approach is architecturally sound and well-scoped.

---

### Issues

**1. Correlated subqueries in `_add_pos_profile_condition` — performance concern**

```sql
AND (
    `tabPricing Rule`.promotional_scheme IS NULL
    OR `tabPricing Rule`.promotional_scheme = ''
    OR NOT EXISTS (
        SELECT 1 FROM `tabPromotion Scheme POS Profile`
        WHERE parent = `tabPricing Rule`.promotional_scheme
    )
    OR EXISTS (
        SELECT 1 FROM `tabPromotion Scheme POS Profile`
        WHERE parent = `tabPricing Rule`.promotional_scheme
        AND pos_profile = %(pos_profile)s
    )
)
```

This adds two correlated subqueries per pricing rule evaluation. ERPNext calls `get_other_conditions` once per item in the cart (inside `get_pricing_rule_for_item`), and each call hits the database with these subqueries. For a cart with 20 items and 50 pricing rules, that's 20 queries each with 2 subqueries scanning `tabPromotion Scheme POS Profile`.

The `tabPromotion Scheme POS Profile` table won't have an index on `(parent, pos_profile)` by default — child tables in Frappe only get an index on `parent`. The `pos_profile` column in the EXISTS subquery will require a full scan of matching parent rows. For small datasets this is fine, but worth noting.

**Suggestion**: Consider adding a composite index, or pre-filtering promotion scheme names and injecting an `IN (...)` clause instead of correlated subqueries.

**2. `get_promotions` filter doesn't apply to standalone Pricing Rules**

The new POS Profile filtering in `get_promotions` (lines 63-82) only filters `schemes` (Promotional Schemes). The standalone `pricing_rules` fetched below (line 106) are returned unfiltered. If a standalone pricing rule should also be restricted to specific POS Profiles, there's no mechanism for that. This may be intentional (standalone rules apply everywhere) but should be documented.

**3. Custom field uses `pos_profiles` fieldname but fixture name is `custom_pos_profiles`**

The custom field JSON defines:
```json
"fieldname": "pos_profiles",
"name": "Promotional Scheme-custom_pos_profiles"
```

In Frappe, custom fields are accessed with the `custom_` prefix (e.g., `doc.custom_pos_profiles`). But the child table `Promotion Scheme POS Profile` uses `parent` to reference the Promotional Scheme — so the query `frappe.get_all("Promotion Scheme POS Profile", filters={"parent": scheme_name})` works regardless of the fieldname. However, if anyone tries to access `scheme_doc.pos_profiles` directly, it won't work — it needs to be `scheme_doc.custom_pos_profiles`. No code in this PR does that, so it's fine, but it's a trap for future developers.

**4. `apply_pricing_rule_on_transaction` also gets filtered — unintended?**

The monkey-patch applies to ALL calls to `get_other_conditions`, which includes both:
- Per-item pricing rules (line 135 in utils.py) — `args` is the item dict merged with doc-level fields
- Transaction-level pricing rules (line 559 in utils.py) — `doc` is the full document

For the per-item path, `pos_profile` propagates from `pricing_args` → `args_copy` correctly. For the transaction-level path, `doc` is the same `pricing_args` dict, so it also has `pos_profile`. This works correctly.

However, if ERPNext internally calls `apply_pricing_rule_on_transaction` from a Sales Invoice that doesn't set `pos_profile` in its args (e.g., standard ERPNext flow, not POS Next), `pos_profile` will be absent, and `_add_pos_profile_condition` returns early without adding the filter — which is correct. No issue here, but the two-path behavior should be documented in the function docstring.

**5. `get_promotions` batch fetch doesn't handle empty `scheme_names`**

```python
scheme_names = [s["name"] for s in schemes]
rows = frappe.get_all(
    "Promotion Scheme POS Profile",
    filters={"parent": ["in", scheme_names]},
    ...
)
```

If `schemes` is empty, `scheme_names` is `[]`, and `frappe.get_all` with `["in", []]` will generate invalid SQL (`parent IN ()`) on some database backends. MariaDB handles it gracefully (returns empty), but it's a code smell. Add an early return.

**6. Removed comment in `__init__.py` was useful**

The PR removes this comment:
```python
# No Frappe hook exists for non-whitelisted module-level functions (override_whitelisted_methods
# only works for @frappe.whitelist() HTTP endpoints, override_doctype_class only for DocType
# classes). This is the standard Python module init approach — runs once at import.
```

This comment explained *why* monkey-patching is used instead of a Frappe hook — which is valuable context for anyone reading the code. The same explanation exists in `pricing_rule.py`'s docstring, so it's not lost, but removing it from the call site (where the "why" matters most) is a minor loss.

---

### Code Quality

**7. `fieldname` in custom field JSON is `pos_profiles` (no `custom_` prefix)**

Frappe auto-generates the actual fieldname as `custom_pos_profiles` based on the `name` field (`Promotional Scheme-custom_pos_profiles`). The `fieldname` property in the JSON says `pos_profiles`, but Frappe overrides this with the name-based convention. This is confusing — the JSON says one thing but the actual field is different. Should set `"fieldname": "custom_pos_profiles"` explicitly for clarity.

**8. The `custom/promotional_scheme.json` approach vs fixtures**

This PR uses Frappe's `{module}/custom/{doctype}.json` convention (with `sync_on_migrate: 1`) instead of the `fixtures` array in `hooks.py`. The rest of the codebase uses `fixtures` for custom fields (e.g., `POS Profile-posa_cash_mode_of_payment`, `Promotional Scheme-pos_only`). Mixed approaches make it harder to track where custom fields are defined. Not a blocker, but worth standardizing.

**9. Missing newline at end of JSON files**

Both `promotion_scheme_pos_profile.json` and `promotional_scheme.json` are missing a trailing newline. Minor but will show up in diffs and linters.

---

### Risk Assessment

| Risk | Level |
|------|-------|
| Correlated subqueries on every item in cart | **Medium** — acceptable for small catalogs, may slow large ones |
| Empty `scheme_names` → invalid SQL | **Low** — MariaDB handles gracefully |
| Mixed fixture approaches | **Low** — maintenance concern only |
| Standalone pricing rules unfiltered | **Low** — likely intentional |

### Verdict

**Looks good overall — minor issues only.** The core mechanism (monkey-patched SQL condition + API-level filtering) is correct and the `pos_profile` propagation through ERPNext's pricing rule engine works for both per-item and transaction-level rules. The main concern is the performance of correlated subqueries at scale — consider a pre-filter approach if POS installations have large pricing rule sets. The other issues are code quality / consistency items. Safe to merge after addressing #5 (empty list guard) and optionally #1 (performance).

engahmed1190 (changes requested):
## Code Review — REQUEST CHANGES

### Summary
Adds per-POS-Profile filtering for Promotional Schemes. A new child table "Promotion Scheme POS Profile" lets admins restrict which POS Profiles a promotion applies to. If left blank, the promotion applies to all profiles.

### What's Good
- **Clean concept**: The child table approach is the right pattern for a many-to-many restriction
- **Request-level caching** in `_add_pos_profile_condition` avoids re-querying the DB for each item in the same transaction
- **`get_promotions` filtering**: Pre-fetches child table rows in bulk (not N+1) and filters correctly
- **New DocType**: `Promotion Scheme POS Profile` is properly defined as `istable: 1`

---

### ISSUE 1 (BLOCKER): Breaks existing `pos_only` logic

The PR restructures `patch_get_other_conditions` and **breaks the existing `pos_only` filtering** that's already on develop.

The current code on develop (lines 67-87 of `pricing_rule.py`) has a working `_patched_get_other_conditions` that:
1. Calls the original function
2. Checks `_has_pos_only_column()`
3. Adds pos_only conditions

The PR replaces this with:
```python
conditions = _original_get_other_conditions(conditions, values, args)
conditions = _add_pos_only_condition(conditions, values, args)
conditions = _add_pos_profile_condition(conditions, values, args)
if not _has_pos_only_column():
    return conditions
```

**Problem**: `_add_pos_only_condition` is called **before** the `_has_pos_only_column()` guard. On sites without POS Next's custom field, this will add `ifnull(tabPricing Rule.pos_only, 0) = 0` to the SQL and crash with "Unknown column". The guard check comes *after* the condition is already appended.

Also, the existing pos_only logic (lines 73-83) is now **duplicated** — it exists both in the new `_add_pos_only_condition` function AND in the original code below the guard check (lines 73-83 are still there). Both will run.

**Fix**: The `_add_pos_only_condition` and `_add_pos_profile_condition` calls must be inside the `_has_pos_only_column()` guard, and the existing inline pos_only logic (lines 73-83) must be removed since it's replaced by `_add_pos_only_condition`.

---

### ISSUE 2 (MEDIUM): `__init__.py` removes doc comments

The PR removes explanatory comments from `__init__.py`:
```python
# No Frappe hook exists for non-whitelisted module-level functions (override_whitelisted_methods
# only works for @frappe.whitelist() HTTP endpoints, override_doctype_class only for DocType
# classes). This is the standard Python module init approach — runs once at import.
```

These comments explain WHY the monkey-patch approach is used. Don't remove documentation that explains non-obvious design decisions.

---

### ISSUE 3 (MEDIUM): Custom field naming — `custom_pos_profiles` vs `pos_profiles`

The `promotional_scheme.json` adds the custom field with `fieldname: "pos_profiles"` but the `name` is `"Promotional Scheme-custom_pos_profiles"`. In Frappe, custom fields should use the `custom_` prefix in their fieldname to avoid collisions with core ERPNext fields. The fieldname should be `custom_pos_profiles`, not `pos_profiles`.

Also, the `pos_only` field is renamed from `"Promotional Scheme-custom_pos_only"` to `"Promotional Scheme-pos_only"` and `insert_after` changed from `"disable"` to `"selling"`. This will cause issues on existing installations that already have the `custom_pos_only` field — the migration path is unclear.

---

### ISSUE 4 (LOW): `frappe.local[cache_key]` uses dynamic attribute setting

```python
cache_key = f"_pos_profile_schemes_{pos_profile}"
if not hasattr(frappe.local, cache_key):
    ...
    frappe.local[cache_key] = {...}
```

`frappe.local` supports `setattr` and `getattr`, but the idiomatic Frappe way is to use `frappe.local.cache` or `frappe.cache()`. Using `setattr` on `frappe.local` directly works but can collide with Frappe internals if the key name happens to match a Frappe attribute.

**Suggestion**: Use `frappe.flags` instead (which is designed for per-request transient data):
```python
if not hasattr(frappe.flags, cache_key):
    frappe.flags[cache_key] = {...}
```

---

### ISSUE 5 (LOW): `pos_profile` not passed in `apply_offers`

The `invoices.py` change adds `"pos_profile": profile.name` to the invoice dict passed to `get_pricing_rules`. This is correct and needed for the POS Profile filtering to work. Good.

However, this means older POS sessions (before this PR) that call `apply_offers` without refreshing the frontend won't pass `pos_profile`. The backend code handles this gracefully (no filter when `pos_profile` is absent), so this is fine.

---

### Verdict
The concept is correct and well-designed, but the implementation has a **critical bug in the monkey-patch ordering** that will crash non-POS-Next sites and duplicate the pos_only filtering. Fix the guard logic, restore the removed comments, and address the custom field naming.

engahmed1190:
@MostafaKadry Please address the blocking issue in the review above — the monkey-patch ordering in `pricing_rule.py` will crash non-POS-Next sites because `_add_pos_only_condition` runs before the `_has_pos_only_column()` guard. Also the existing inline pos_only logic (lines 73-83) is duplicated since `_add_pos_only_condition` does the same thing.

Quick fix outline:
1. Move both `_add_pos_only_condition` and `_add_pos_profile_condition` calls **inside** the `_has_pos_only_column()` guard
2. Remove the duplicate inline pos_only logic (lines 73-83) since it's now handled by `_add_pos_only_condition`
3. Restore the removed comments in `__init__.py`
4. Use `custom_pos_profiles` as the fieldname (not `pos_profiles`) to follow Frappe custom field conventions

PR #181, opened 2026-03-09
PN-17: Credit return to wallet and auto-create wallet on customer insert

MohamedAliSmk:
- Add create_wallet_on_customer_insert hook for Customer after_insert
- Add force_create parameter to get_or_create_wallet for manual wallet creation
- Add credit_return_to_wallet function for crediting return amounts to customer wallet
- Improve wallet transaction GL entries with party info for Receivable/Payable accounts
- Improve error handling for broken wallet transactions (recover draft, cancel broken)
- Refactor payment account assignment into _set_payment_accounts helper (bug fix)
- Add add_to_customer_balance flag in ReturnInvoiceDialog
- Improve wallet reversal and credit error handling in submit_invoice

### DIFF
--- a/POS/src/components/sale/ReturnInvoiceDialog.vue
+++ b/POS/src/components/sale/ReturnInvoiceDialog.vue
@@ -1052,6 +1052,8 @@ const createReturnResource = createResource({
 				// Link to original invoice item row for accurate return tracking in ERPNext
 				sales_invoice_item: item.name,
 			})),
+			// Flag to indicate return amount should be added to customer credit balance
+			add_to_customer_balance: addToCustomerCredit.value,
 			// Payment amounts are negative for refunds
 			// If addToCustomerCredit is true, send empty payments array so outstanding stays negative
 			// This negative outstanding becomes customer credit balance
--- a/pos_next/api/invoices.py
+++ b/pos_next/api/invoices.py
@@ -340,9 +340,9 @@ def get_payment_account(mode_of_payment, company):
 def _set_payment_accounts(payments, company):
     """Set the account for each payment entry that is missing one.
 
-    Uses BaseDocument.set() for assignment which directly writes to __dict__,
-    avoiding potential issues with attribute assignment on child-table Document
-    objects (e.g. 'SalesInvoicePayment' object does not support item assignment).
+    Handles both Document objects (from invoice_doc.payments) and plain dicts
+    (from frontend data).  Document objects use BaseDocument.set() which writes
+    directly to __dict__, while plain dicts use normal key assignment.
     """
     if not payments or not company:
         return
@@ -354,12 +354,18 @@ def _set_payment_accounts(payments, company):
         try:
             account_info = get_payment_account(mode_of_payment, company)
             if account_info:
-                payment.set("account", account_info.get("account"))
+                account = account_info.get("account")
+                if hasattr(payment, "set") and callable(payment.set):
+                    payment.set("account", account)
+                else:
+                    payment["account"] = account
         except Exception as e:
             frappe.log_error(
                 f"Failed to get payment account for {mode_of_payment}: {e}",
                 "Payment Account Lookup",
             )
+
+
 # ==========================================
 # Stock Validation Functions
 # ==========================================
@@ -1353,6 +1359,7 @@ def submit_invoice(invoice=None, data=None):
         invoice_doc.submit()
         invoice_submitted = True
         # Handle wallet transaction reversal for returns
+        wallet_reversal_ok = False
         if invoice_doc.get("is_return") and invoice_doc.get("return_against"):
             from pos_next.pos_next.doctype.wallet_transaction.wallet_transaction import reverse_wallet_transactions_for_return
             try:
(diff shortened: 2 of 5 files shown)


engahmed1190:
## Deep Code Review — PR #181

**Files changed**: 5 | **+355 / -166**

---

### Critical Issues

**1. Wrong import path in `credit_return_to_wallet` — will crash or misbehave at runtime**

In `wallet_transaction.py:340`:
```python
from pos_next.pos_next.doctype.wallet.wallet import get_or_create_wallet
```

This imports from `wallet.py` (the doctype), which has a **different signature**: `get_or_create_wallet(customer, company)` — no `force_create` param, no POS settings logic, and no conditional return of `None`. The API version at `pos_next.api.wallet` is the one with `force_create`. This will work but **won't respect `force_create`** semantics and will unconditionally create wallets, which may be unintended. More importantly, the wallet doctype version and the API version have diverged — this is a maintenance landmine.

**2. Missing blank line before section comment**

In `invoices.py`, after `_set_payment_accounts` there's no blank line before the `# ==========================================` comment block. Missing the standard two blank lines between top-level definitions. Indicates the PR wasn't linted.

**3. `_set_payment_accounts` uses `.set()` but will fail on plain dicts**

The docstring claims it uses `BaseDocument.set()` to avoid attribute assignment issues, but the `payment` objects could be either dicts (from frontend data) or Document objects (from `invoice_doc.payments`). The function calls `payment.get()` which works on both, but `payment.set("account", ...)` will **fail on plain dicts** — dicts don't have a `.set()` method. The original inline code in `submit_invoice` used `payment.account = ...` (attribute assignment). This refactor introduces a regression for any code path where payments are dicts.

**4. `frappe.db.commit()` mid-transaction breaks atomicity**

In `credit_return_to_wallet` (~line 387):
```python
existing_transaction.cancel()
frappe.db.commit()
```
This is called from within `submit_invoice`, which has its own transaction/commit logic. Calling `frappe.db.commit()` in the middle breaks the atomicity of the outer transaction. If the subsequent re-creation fails, the cancellation is already committed and can't be rolled back.

**5. `add_to_customer_balance` placed after `except` — risk of double-crediting**

In `invoices.py`, the wallet credit block is placed **after** the `except` for wallet reversal, but still inside the `if invoice_doc.get("is_return")` block. If wallet reversal fails, execution continues to `credit_return_to_wallet`. This means both a reversal failure error AND a new credit could happen, potentially double-crediting the customer.

---

### Logic Concerns

**6. Loyalty program `min_spent` threshold — hidden business logic change**

The new code in `reverse_wallet_transactions_for_return` fetches `min_spent` from the loyalty program and cancels the entire wallet transaction if the remaining invoiced amount drops below the threshold. This is a business logic change beyond what the PR title suggests. It also only checks one `Loyalty Program Collection` row — if there are multiple tiers, `frappe.db.get_value` returns just the first one arbitrarily.

**7. `create_wallet_on_customer_insert` hook fires for ALL customers**

Every customer created in the system (not just POS customers) will get a wallet auto-created. This could create thousands of unnecessary Wallet documents for businesses that use ERPNext beyond POS. There's no opt-out setting — the hook ignores `auto_create_wallet` from POS settings entirely.

**8. Inconsistent idempotency guards**

The debit idempotency in `reverse_wallet_transactions_for_return` uses:
```python
frappe.db.exists("Wallet Transaction", {
    "reference_name": return_invoice, "transaction_type": "Debit", "docstatus": ["!=", 2]
})
```
But the credit idempotency in `credit_return_to_wallet` adds `"source_type": "Refund"` to the filter. Inconsistent guard strategies across the two functions.

---

### Performance

**9. Extra DB queries in GL entry building**

`build_gl_entries` now calls `frappe.get_cached_value("Account", source_account, "account_type")` on every invocation. For the Debit path, it caches in `self._source_account_type`, but inconsistently — the Credit path doesn't cache. Minor since `get_cached_value` uses Redis cache, but the inconsistency suggests copy-paste without thought.

**10. `credit_return_to_wallet` fetches full Sales Invoice document unnecessarily**

`frappe.get_doc("Sales Invoice", return_invoice)` loads the entire document with all child tables. The function only needs `customer`, `company`, `grand_total`, `is_return`, and `return_against`. Should use `frappe.db.get_value` for just those fields.

---

### Code Quality

**11. Mixed indentation in `reverse_wallet_transactions_for_return`**

The idempotency check block uses spaces while the rest uses tabs:
```python
    existing = frappe.db.exists("Wallet Transaction", {
    "reference_name": return_invoice, ...
    })
```
Will cause issues with strict linters.

**12. Lazy imports from sibling doctype**

`credit_return_to_wallet` imports `get_or_create_wallet` inside the function body from a sibling doctype. No circular dependency risk here — should be at the top of the file.

---

### Risk Assessment

| Risk | Level |
|------|-------|
| Runtime crash from `.set()` on dicts | **High** — will crash on dict payments |
| `frappe.db.commit()` mid-transaction | **High** — breaks atomicity |
| Wrong import path for `get_or_create_wallet` | **Medium** — uses wrong function |
| Double credit on reversal failure | **Medium** — financial data integrity |
| Wallet created for all customers | **Medium** — unwanted side effect |
| Loyalty `min_spent` one-tier assumption | **Low** — edge case |

### Verdict

**Not ready to merge.** The PR has at least 2-3 issues that will cause runtime errors (`_set_payment_accounts` dict crash, `frappe.db.commit()` breaking atomicity) and a business logic concern (wallets for all customers). The `credit_return_to_wallet` function needs the import path fixed and the idempotency recovery reworked without mid-transaction commits. Recommend addressing the critical items and re-reviewing.

PR #183, opened 2026-03-10
feat: allow searching items by item group in POS search

abinandabinand21-sudo:
### Summary
This PR enhances the POS item search functionality by allowing users to search items using the **Item Group** name directly from the POS search bar.

### Problem
When there are many item groups, scrolling and selecting the correct item group filter becomes time-consuming for the cashier.

### Before
Previously, the POS search bar only supported searching by **item code**, **item name**, or **barcode**.  
If a cashier wanted to find items belonging to a specific **Item Group**, they had to manually scroll through the available item group filters and select the desired group.

This process becomes inconvenient and slower when there are many item groups available in the POS interface.

### Solution
The POS search functionality has been extended to include **Item Group names** as part of the searchable fields.

Users can now type the name of an item group directly in the search bar to quickly display items belonging to that group.

### After
Screencastfrom2026-03-1014-02-50-ezgif com-video-to-gif-converter


engahmed1190:
Wonderful @abinandabinand21-sudo

PR #206, opened 2026-03-25
fix: restore coupon usage when invoices are cancelled

### LINKED ISSUE #195: [Bug]: Cancelling an invoice does not restore coupon usage
### Bug Description
Coupon usage is incremented during sale processing, but invoice cancellation does not restore that usage.

There is a decrement helper in code, but it does not appear to be wired into a reliable invoice cancel path.

As a result, a cancelled invoice can permanently consume a limited-use coupon or gift-card-style coupon.

### Expected Behavior
If an invoice that consumed a coupon is cancelled successfully, the coupon usage count should be reversed.

### Actual Behavior
Cancelled invoices continue to count against coupon usage limits.

### Steps to Reproduce
1. Create a `POS Coupon` with a limited usage rule
2. Use the coupon on a sale
3. Cancel the submitted invoice
4. Check the coupon usage counter / reusability

### Impact
- coupon balances and usage counts drift away from successful sales
- gift-card and limited-use coupons can be exhausted incorrectly
- operators cannot reliably reuse a coupon after a cancelled sale

### Suggested Fix
- reverse coupon usage from a reliable invoice cancellation hook
- keep coupon mutations within the same transaction boundaries as the sale lifecycle

### Notes
This appears separate from the one-use doctype mismatch issue. Even when coupon usage is recorded, cancellation does not properly unwind it.

my-dev-jour:
## Summary
- add a Sales Invoice on_cancel hook to restore POS coupon usage after successful cancellation
- keep credit JE cleanup on before_cancel and realtime updates on on_cancel
- add regression tests for coupon rollback hook behavior

## Testing
- python3 -m py_compile pos_next/api/sales_invoice_hooks.py pos_next/hooks.py pos_next/api/test_sales_invoice_hooks.py
- full Frappe unit tests were not run in this workspace because the frappe module is not installed here

Closes #195

### DIFF
--- a/pos_next/api/sales_invoice_hooks.py
+++ b/pos_next/api/sales_invoice_hooks.py
@@ -141,3 +141,32 @@ def before_cancel(doc, method=None):
 			alert=True,
 			indicator="orange"
 		)
+
+
+def on_cancel(doc, method=None):
+	"""
+	On Cancel hook for Sales Invoice.
+	Roll back coupon usage only after the invoice has cancelled successfully.
+
+	Args:
+		doc: Sales Invoice document
+		method: Hook method name (unused)
+	"""
+	rollback_coupon_usage(doc)
+
+
+def rollback_coupon_usage(doc):
+	"""Restore coupon usage for cancelled invoices that consumed a POS coupon."""
+	coupon_code = doc.get("coupon_code")
+	if not coupon_code or not frappe.db.table_exists("POS Coupon"):
+		return
+
+	try:
+		from pos_next.pos_next.doctype.pos_coupon.pos_coupon import decrement_coupon_usage
+
+		decrement_coupon_usage(coupon_code)
+	except Exception as e:
+		frappe.log_error(
+			title="Coupon Usage Rollback Failed",
+			message=f"Invoice: {doc.name}, Coupon: {coupon_code}, Error: {str(e)}\n{frappe.get_traceback()}",
+		)
--- a/pos_next/api/sales_invoice_hooks.py
+++ b/pos_next/api/sales_invoice_hooks.py
@@ -157,6 +157,9 @@ def on_cancel(doc, method=None):
 
 def rollback_coupon_usage(doc):
 	"""Restore coupon usage for cancelled invoices that consumed a POS coupon."""
+	if doc.get("is_return"):
+		return
+
 	coupon_code = doc.get("coupon_code")
 	if not coupon_code or not frappe.db.table_exists("POS Coupon"):
 		return
--- a/pos_next/api/test_sales_invoice_hooks.py
+++ b/pos_next/api/test_sales_invoice_hooks.py
@@ -12,7 +12,10 @@ class TestSalesInvoiceHooks(unittest.TestCase):
 	@patch("pos_next.pos_next.doctype.pos_coupon.pos_coupon.decrement_coupon_usage")
 	def test_rollback_coupon_usage_for_cancelled_invoice(self, mock_decrement, mock_db):
 		doc = Mock()
-		doc.get.return_value = "SAVE10"
+		doc.get.side_effect = lambda key, default=None: {
+			"is_return": 0,
+			"coupon_code": "SAVE10",
+		}.get(key, default)
 		doc.name = "ACC-SINV-0001"
 		mock_db.table_exists.return_value = True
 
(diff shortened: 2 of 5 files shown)


engahmed1190:
### Code review

Found 1 issue:

1. `rollback_coupon_usage` has no `is_return` guard -- cancelling a return (credit note) invoice will decrement the coupon usage counter even though no usage was ever incremented for the return. Return invoices inherit `coupon_code` from the original invoice via ERPNext's `make_sales_return`. This can drive the `used` counter below its correct value, allowing single-use coupons to be reused. Add an early return for `doc.get("is_return")`.

https://github.com/BrainWise-DEV/POSNext/blob/1e9b0c3cd9aec92bea887b2853bd6382ed02c991/pos_next/api/sales_invoice_hooks.py#L158-L167

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

my-dev-jour:
Addressed. I added an early return for `doc.get("is_return")` in `rollback_coupon_usage()` so return/credit-note cancellations do not decrement coupon usage. I also added regression coverage for both the return-invoice path and the missing-coupon-table path.

engahmed1190 (changes requested):
## Code Review — REQUEST CHANGES ❌

### Summary
Adds `on_cancel` hook for Sales Invoice to restore POS coupon usage via `decrement_coupon_usage`. Good edge case handling for return invoices and missing coupon table.

### Test Results
**4/5 tests pass**, 1 fails. `test_rollback_coupon_usage_skips_when_coupon_table_missing` fails because the `frappe.db.table_exists` mock doesn't properly intercept the Werkzeug LocalProxy when running under `bench execute`.

### Correctness

**Logic is correct** — `on_cancel` delegates to `rollback_coupon_usage`, which:
- Skips return invoices (correct — returns copy `coupon_code` but shouldn't affect usage)
- Skips when coupon table doesn't exist (graceful degradation)
- Wraps `decrement_coupon_usage` in try/except to not block cancellation

**hooks.py change**: Correctly converts `on_cancel` from string to list. Good.

### Issue: `frappe.db.commit()` in `decrement_coupon_usage` (BLOCKING)

`decrement_coupon_usage` internally calls `frappe.db.commit()`. When called from an `on_cancel` hook, this commits the coupon decrement *mid-transaction*. If subsequent hooks or Frappe's own cancel logic fails:
- The coupon usage count is already decremented (committed)
- But the invoice cancellation rolls back
- Result: coupon usage is decremented without a corresponding cancellation — data inconsistency

This is the same issue PR #207 fixes for the *increment* path. The decrement path needs the same treatment.

### Required Fix
Remove `frappe.db.commit()` from `decrement_coupon_usage`, or create an atomic version analogous to `consume_coupon_usage` from PR #207. The coupon decrement must participate in the same transaction as the invoice cancellation.

### Additional Suggestions (non-blocking)
1. **Idempotency**: No guard against double-decrement if cancel is retried. `decrement_coupon_usage` checks `coupon.used > 0` before decrementing (floor at zero), which is acceptable but not ideal
2. **Missing test**: No test verifying behavior when `decrement_coupon_usage` raises an exception (the try/except path). Assert `frappe.log_error` is called
3. **Locking**: No `SELECT FOR UPDATE` on the decrement path — two simultaneous cancellations using the same coupon could race. Low risk (cancellations are rare) but worth noting for symmetry with #207
4. **Fix the test failure**: Mock `frappe.db` at `frappe.local.db` level or restructure to avoid Werkzeug LocalProxy issues

### Verdict
The overall approach is correct. Remove the `frappe.db.commit()` from `decrement_coupon_usage` to maintain transactional integrity with the document's cancel operation, then this is ready to merge.

my-dev-jour:
Addressed the new blocker in branch commit e9db0eb. I replaced the cancel-side coupon rollback with a transactional row-lock update path:  now delegates to a new  helper that uses  and  without any internal . That keeps coupon rollback in the same transaction as invoice cancellation instead of committing it independently. I also added focused tests for the no-commit rollback path and for re-raising/logging failures.

my-dev-jour:
Addressed the new blocker in branch commit e9db0eb.

I replaced the cancel-side coupon rollback with a transactional row-lock update path:
- `decrement_coupon_usage()` now delegates to a new `rollback_coupon_usage()` helper
- the helper uses `SELECT ... FOR UPDATE`
- it updates usage through `frappe.db.set_value(..., update_modified=False)`
- it does not call any internal `frappe.db.commit()`

That keeps coupon rollback in the same transaction as invoice cancellation instead of committing it independently.

I also added focused tests for:
- the no-commit rollback path
- re-raising and logging failures

PR #207, opened 2026-03-25
fix: make coupon consumption atomic during submit

### LINKED ISSUE #198: [Bug]: Coupon validation and consumption are race-prone under concurrent checkout
### Bug Description
Coupon validation checks current usage first, and the usage increment happens later in a separate step with no lock or atomic conditional update.

That means two concurrent checkouts can both pass validation before either request increments usage.

### Expected Behavior
Coupon consumption should be atomic for limited-use and gift-card-style coupons.

### Actual Behavior
Validation and consumption are split into separate non-atomic steps.

### Steps to Reproduce
1. Create a coupon with a low usage limit, for example `maximum_use = 1`
2. Start two checkout sessions at nearly the same time
3. Apply the same coupon in both sessions
4. Submit both checkouts concurrently

### Impact
- supposedly single-use coupons can be oversubscribed
- gift-card-style balances can be consumed by more than one checkout under concurrency
- coupon limits become unreliable on busy POS deployments

### Suggested Fix
- lock the coupon row or use an atomic conditional update
- revalidate the current usage inside the same transaction immediately before incrementing

### Notes
This is a transaction-safety problem and appears independent of the one-use doctype mismatch bug.

my-dev-jour:
## Summary
- lock the coupon row and revalidate it before consuming usage during invoice submit
- consume coupon usage inside the submit transaction instead of committing from a helper
- add regression tests for one-use usage counting and atomic consume behavior

## Testing
- python3 -m py_compile pos_next/pos_next/doctype/pos_coupon/pos_coupon.py pos_next/api/invoices.py pos_next/pos_next/doctype/pos_coupon/test_pos_coupon.py

Closes #198

### DIFF
--- a/pos_next/api/invoices.py
+++ b/pos_next/api/invoices.py
@@ -1302,14 +1302,13 @@ def submit_invoice(invoice=None, data=None):
         if coupon_code:
             # Increment usage counter for POS Coupon
             if frappe.db.table_exists("POS Coupon"):
-                try:
-                    from pos_next.pos_next.doctype.pos_coupon.pos_coupon import increment_coupon_usage
-                    increment_coupon_usage(coupon_code)
-                except Exception as e:
-                    frappe.log_error(
-                        title="Failed to increment coupon usage",
-                        message=f"Coupon: {coupon_code}, Error: {str(e)}"
-                    )
+                from pos_next.pos_next.doctype.pos_coupon.pos_coupon import consume_coupon_usage
+
+                consume_coupon_usage(
+                    coupon_code,
+                    customer=invoice_doc.customer,
+                    company=invoice_doc.company,
+                )
 
         # Auto-set batch numbers for returns
         _auto_set_return_batches(invoice_doc)
--- a/pos_next/pos_next/doctype/pos_coupon/pos_coupon.py
+++ b/pos_next/pos_next/doctype/pos_coupon/pos_coupon.py
@@ -176,18 +176,44 @@ def apply_coupon_discount(coupon, cart_total, net_total=None):
     }
 
 
-def increment_coupon_usage(coupon_code):
-    """Increment the usage counter for a coupon"""
+def consume_coupon_usage(coupon_code, customer=None, company=None):
+    """Atomically validate and increment coupon usage within the current transaction."""
+    coupon_code = coupon_code.upper()
+    locked_coupon = frappe.db.sql(
+        """
+        SELECT name
+        FROM `tabPOS Coupon`
+        WHERE coupon_code = %s
+        FOR UPDATE
+        """,
+        (coupon_code,),
+        as_dict=True,
+    )
+
+    if not locked_coupon:
+        frappe.throw(_("Sorry, this coupon code does not exist"))
+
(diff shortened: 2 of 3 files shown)


engahmed1190:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

engahmed1190:
Fix Conflicts

my-dev-jour:
Conflicts resolved. I rebased the branch onto current `develop`, kept the already-merged one-use doctype logic from #205, and retained the atomic coupon-consumption changes plus regression coverage.

engahmed1190 (commented):
## Code Review — APPROVE with comments ✅

### Summary
Makes coupon consumption atomic during invoice submission using `SELECT FOR UPDATE` row locking. Prevents race conditions where two concurrent submissions both pass the "usage < max" check.

### Test Results
**1/3 tests pass**, 2 fail. The failures are caused by a Python 3.10 `unittest.mock` issue — when `frappe.db.count` is mocked at module level while frappe is initialized, the mock returns a coroutine instead of an int. The new `consume_coupon_usage` test passes.

**Fix**: The two failing tests are from PR #205 (already merged) and pre-exist this PR. They need `mock_db.count.return_value = <int>` instead of the current `side_effect` pattern, or should use `patch.object(frappe.local, 'db', ...)`.

### Correctness

**Atomic pattern is correct:**
- `SELECT ... FOR UPDATE` locks the coupon row, blocking concurrent transactions
- `check_coupon_code()` re-validates within the same transaction that holds the lock
- `frappe.db.set_value` updates the counter atomically
- No deadlock risk — single-row lock per coupon, released at transaction commit
- `update_modified=False` is correct — counter updates shouldn't trigger modified timestamps

**Behavior change (intentional and correct):**
The old code wrapped `increment_coupon_usage` in a try/except that swallowed errors — a failed coupon increment would still allow the invoice to submit. The new code propagates the failure, blocking invoice submission. This is the correct behavior: an invoice should not submit with a coupon if we can't verify/track usage.

**`increment_coupon_usage` backward-compat wrapper:**
Good — logs the error AND re-raises, so callers using the old API get both logging and proper error propagation.

1. **Add negative-path tests**: test what happens when `FOR UPDATE` returns empty (coupon not found) and when `check_coupon_code` returns `valid: False`
2. **Double `_()` wrapping**: `frappe.throw(_(coupon_result.get("msg", "Invalid coupon code")))` — the messages from `check_coupon_code` are already wrapped in `_()`. The outer `_()` is redundant (harmless but unnecessary)
3. **Redundant fetch**: `check_coupon_code` does `frappe.get_doc()` internally which re-reads the coupon. The `FOR UPDATE` query already has the coupon `name`. Minor optimization would be to pass the locked data to a validation function, but not a correctness issue
4. **Asymmetry with decrement**: `decrement_coupon_usage` (used by PR #206) still uses `frappe.get_doc` + `db_set` + `db.commit()` without locking. Consider updating for symmetry in a follow-up

### Verdict
Core logic is correct and materially improves concurrency safety. Fix the test infrastructure issue and this is ready to merge.

PR #208, opened 2026-03-25
fix: make customer search and loyalty assignment context-aware

### LINKED ISSUE #202: [Bug]: Customer search API ignores search_term and returns browse results
### Bug Description
The customer lookup API accepts `search_term`, but the server-side query does not actually apply it.

The endpoint returns a limited alphabetical list instead of filtering by the operator's typed input.

### Expected Behavior
Customer search should filter results using the provided search term across intended lookup fields such as:
- customer ID
- customer name
- mobile number
- other configured search fields

### Actual Behavior
`search_term` is effectively ignored, so the endpoint behaves like a paged browse list rather than a real search API.

### Steps to Reproduce
1. Populate enough customers that the desired customer is not in the first page alphabetically
2. Search for that customer in POSNext using a distinctive term
3. Inspect the returned results

### Impact
- operators cannot reliably find the intended customer on large datasets
- inline POS customer search becomes slow and error-prone
- typed search text does not map to actual server-side filtering

### Suggested Fix
- apply search predicates on the server before limit/order
- keep pagination only after the query performs real filtering

### Notes
This affects day-to-day operator usability directly and is easy to miss on small demo datasets.

### LINKED ISSUE #203: [Bug]: Loyalty auto-assignment uses the first enabled POS Settings row instead of the active context
### Bug Description
Customer loyalty auto-assignment selects the first enabled POS Settings record with a default loyalty program, instead of resolving from the actual POS Profile / company context that created the customer.

### Expected Behavior
The loyalty program assigned to a new customer should match the company/profile context of the current POS session.

### Actual Behavior
The selection is global and order-dependent rather than context-aware.

### Steps to Reproduce
1. Configure multiple POS settings/profile contexts with different default loyalty programs
2. Create customers from different POS contexts
3. Compare the assigned loyalty program with the context that created each customer

### Impact
- wrong loyalty program can be assigned in multi-profile or multi-company setups
- customer creation behavior becomes inconsistent across contexts
- downstream loyalty earning/redemption rules may be wrong from the start

### Suggested Fix
- resolve default loyalty program from the active POS Profile/company context
- avoid using the first enabled settings row as a global fallback

### Notes
This is primarily a context-resolution bug and becomes much more visible on multi-company or multi-profile deployments.

my-dev-jour:
## Summary
- apply real search_term filtering to customer lookup instead of returning browse-only results
- make loyalty auto-assignment use explicit company/profile context instead of the first enabled POS Settings row
- avoid assigning a loyalty program when the company context is ambiguous
- add regression tests for customer search filters and loyalty context resolution

## Testing
- python3 -m py_compile pos_next/api/customers.py pos_next/api/test_customers.py

Closes #202
Closes #203

### DIFF
--- a/POS/src/components/sale/CreateCustomerDialog.vue
+++ b/POS/src/components/sale/CreateCustomerDialog.vue
@@ -332,17 +332,14 @@ const updateTerritoryFromCountry = () => {
 // =============================================================================
 
 const createCustomerResource = createResource({
-	url: "frappe.client.insert",
+	url: "pos_next.api.customers.create_customer",
 	makeParams: () => ({
-		doc: {
-			doctype: "Customer",
-			customer_name: customerData.value.customer_name,
-			customer_type: "Individual",
-			customer_group: customerData.value.customer_group || __("Individual"),
-			territory: customerData.value.territory || __("All Territories"),
-			mobile_no: customerData.value.mobile_no || "",
-			email_id: customerData.value.email_id || "",
-		},
+		customer_name: customerData.value.customer_name,
+		mobile_no: customerData.value.mobile_no || "",
+		email_id: customerData.value.email_id || "",
+		customer_group: customerData.value.customer_group || __("Individual"),
+		territory: customerData.value.territory || __("All Territories"),
+		pos_profile: props.posProfile,
 	}),
 	onSuccess: (data) => {
 		showSuccess(__("Customer {0} created successfully", [data.customer_name]))
--- a/pos_next/api/customers.py
+++ b/pos_next/api/customers.py
@@ -28,6 +28,7 @@ def get_customers(search_term="", pos_profile=None, limit=20, modified_since=Non
         )
 
         filters = {}
+        or_filters = []
 
         # Filter by POS Profile customer group if specified
         if pos_profile:
@@ -45,10 +46,21 @@ def get_customers(search_term="", pos_profile=None, limit=20, modified_since=Non
             # Full fetch: only active customers
             filters["disabled"] = 0
 
+        search_term = (search_term or "").strip()
+        if search_term:
+            like_term = f"%{search_term}%"
+            or_filters = [
+                ["Customer", "name", "like", like_term],
+                ["Customer", "customer_name", "like", like_term],
+                ["Customer", "mobile_no", "like", like_term],
+                ["Customer", "email_id", "like", like_term],
+            ]
+
         customer_limit = limit if limit not in (None, 0) else frappe.db.count("Customer", filters)
         result = frappe.get_all(
             "Customer",
             filters=filters,
+            or_filters=or_filters or None,
             fields=["name", "customer_name", "mobile_no", "email_id", "disabled"],
             limit=customer_limit,
             order_by="customer_name asc",
@@ -62,7 +74,15 @@ def get_customers(search_term="", pos_profile=None, limit=20, modified_since=Non
 
 
 @frappe.whitelist()
-def create_customer(customer_name, mobile_no=None, email_id=None, customer_group="Individual", territory="All Territories", company=None):
+def create_customer(
+    customer_name,
+    mobile_no=None,
+    email_id=None,
+    customer_group="Individual",
+    territory="All Territories",
+    company=None,
+    pos_profile=None,
+):
     """
     Create a new customer from POS.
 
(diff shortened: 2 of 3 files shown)


engahmed1190:
### Code review

Found 1 issue:

1. **Loyalty auto-assignment is silently broken for customers created via the POS UI.** The PR adds `frappe.flags` context-passing in `create_customer()`, but the actual POS frontend (`CreateCustomerDialog.vue` line 335) creates customers via `frappe.client.insert` -- it never calls `pos_next.api.customers.create_customer`. This means `frappe.flags.pos_next_customer_company` and `frappe.flags.pos_next_customer_pos_profile` are never set for the normal POS flow. The `form_dict` fallback in `_get_customer_assignment_context()` also won't help because `frappe.client.insert` nests fields inside `form_dict.doc`, not at the top level where `form_dict.get("company")` looks. The net result is `_get_customer_assignment_context()` returns `(None, None)`, causing `get_default_loyalty_program_from_settings(None, None)` to return `None` -- no loyalty program is ever assigned for POS-created customers.

https://github.com/BrainWise-DEV/POSNext/blob/35f43d41c016910943df5b6080b830c37e286299/pos_next/api/customers.py#L122-L131

Relevant frontend call site that bypasses `create_customer()`:

https://github.com/BrainWise-DEV/POSNext/blob/1f1a3862dbd16b07fe2f66e1f75f94f21d17cee0/POS/src/components/sale/CreateCustomerDialog.vue#L334-L336

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

my-dev-jour:
I reworked the loyalty part based on the actual POS customer-create flow. Staging validation showed the reviewer was right: with a temporary loyalty program configured, plain `frappe.client.insert` produced a customer with no loyalty program, while `pos_next.api.customers.create_customer(..., pos_profile="Main Counter")` assigned the expected program. The updated branch now switches the POS dialog to the explicit POSNext API and adds regression coverage for the backend `create_customer` path.

engahmed1190 (commented):
## Code Review — APPROVE with suggestions ✅

### Summary
Three related fixes: (1) actual search filtering for customer lookup, (2) context-aware loyalty program assignment, (3) frontend routing through proper POS API.

### Test Results
**4/5 tests pass**, 1 fails. `test_get_default_loyalty_program_from_settings_uses_explicit_pos_profile` fails because `frappe.db.get_value` returns a coroutine when mocked at module level in Python 3.10 with frappe initialized (Werkzeug LocalProxy issue).

### Correctness

**Customer search filtering:**
- `or_filters` with LIKE matching on `name`, `customer_name`, `mobile_no`, `email_id` is the standard Frappe pattern
- `or_filters or None` guard correctly avoids passing empty list when no search term
- **No SQL injection risk**: Frappe's `get_all()` with `or_filters` goes through `DatabaseQuery.prepare_filter_condition()` which escapes string values via `frappe.db.escape()`

**Loyalty program context:**
- `pos_profile` lookup path: direct `frappe.db.get_value` by profile — correct and efficient
- `company`-only path: resolves all POS Settings, checks profile companies, returns program only when unambiguous — correct and safe
- Ambiguity check (returning `None` when 2 different programs exist for same company) is intentionally conservative

**`frappe.flags` pattern:**
- Safe: flags are per-request (re-initialized in `frappe.init()`)
- The `try/finally` cleanup in `create_customer` ensures no leakage even if `insert()` raises
- No risk of cross-request contamination

**Frontend:**
- Switching from `frappe.client.insert` to `pos_next.api.customers.create_customer` is a security improvement — generic endpoint replaced with validated, permission-checked custom endpoint

1. **Add logging on ambiguity skip**: When `get_default_loyalty_program_from_settings` returns `None` because of ambiguous programs, there's no indication why. A `frappe.logger().info("Multiple loyalty programs found for company {company}, skipping auto-assignment")` would help administrators diagnose "why didn't my customer get a loyalty program?"
2. **Fix the failing test**: Use `mock_db.get_value.return_value = "LOYALTY-A"` explicitly or mock at `frappe.local.db` level
3. **Missing test cases**: `company`-only path when all profiles agree (should return the single program), and `company=None, pos_profile=None` (should return `None`)
4. **`customer_limit` with `or_filters`**: When `limit` is None, falls back to `frappe.db.count("Customer", filters)` which doesn't account for `or_filters`. This just means the limit is higher than needed — harmless but worth noting

### Verdict
Good improvements across all three areas. Fix the test failure and consider adding the ambiguity logging, then merge.

engahmed1190 (approved):
Confirming approval — customer search filtering and context-aware loyalty assignment are correct. The one failing test is a mock infrastructure issue, not a code bug. Safe to merge.

Issue #191, opened 2026-03-19
[Bug]: incorrect Paid and Outstanding amounts


gp381:
### Issue Search Confirmation

- [x] I have searched existing open and closed issues and this is not a duplicate

### Bug Description

I am using ERPNext 16, with POSNext **develop** branch

I have 2 different currencies in my system
the company default currency is : **SAR**
the selling currency is : **YER**
exchange rate from YER to SAR  = 0.0023529412 SAR
exchange rate from SAR to YER  = 425 YER

when I have an Unpaid, Overdue, Partial Paid invoices.
the paid and outstanding amount is incorrect
like the following image

<img width="2148" height="416" alt="Image" src="https://github.com/user-attachments/assets/55f7f16d-cda3-4ff0-8ca6-b5635b537398" />

the current result is
Total Amount : YER 96,700.000000000
Paid : YER 96,472.470702000
Outstanding : YER 227.529298000

the outstanding is showing in SAR amount with YER currency code
if you multiply Outstanding(227.529298000) * SAR_to_YER(425) = YER 96,700.00

the paid amount looks like `Total Amount - Outstanding`

the expected result should be

Total Amount : YER 96,700.000000000
Paid : YER 0
Outstanding : YER 96,700.000000000

the outstanding should be showing in YER amount and YER currency code

the paid amount should be calculated depending on the actual paid amount in the system just in case the paid amount is more than the Total amount
------
this is another invoice with partial paid status
the invoice details are
The invoice Total amount : 10,000 YER
the actual paid amount : 5,000 YER
the outstanding amount : 5,000 YER

<img width="2160" height="667" alt="Image" src="https://github.com/user-attachments/assets/84392c1a-1f4b-4033-9ea2-6266c524e9ed" />

the current result is

Total Amount : YER 10,000.000000000
Paid : YER 9,988.235300000
Outstanding : YER 11.764700000

Payment History : YER 11.764700000

the expected result should be

Total Amount : YER 10,000.000000000
Paid : YER 5,000.000000000
Outstanding : YER 5,000.000000000

Payment History : YER 5,000.000000000

### Steps to Reproduce

1. setup a company with SAR as default currency.
2. setup the selling default currency to YER.
3. add exchange rate for SAR to YER = 425.
4. add exchange rate for YER to SAR = 0.0023529412.
5. add an item with Standard Selling in YER = 10,000 YER.
6. go to POSNext and sell that item **On Account** watch the first issue
[...]

PR #209, opened 2026-03-25
fix: rebase POS sales invoice GL entries on ERPNext 16

my-dev-jour:
## Summary
- rebase POSNext's Sales Invoice POS GL override on the ERPNext 16 core make_pos_gl_entries flow
- restore ERPNext 16's against_voucher guard and transaction-currency fields
- keep the wallet-specific party_type/party behavior on payment-mode accounts

## Why
POSNext's override had drifted from ERPNext 16 core and could produce incorrect paid/outstanding behavior on submitted POS invoices.

## Testing
- python3 -m py_compile pos_next/overrides/sales_invoice.py

Refs #191

### DIFF
--- a/pos_next/overrides/sales_invoice.py
+++ b/pos_next/overrides/sales_invoice.py
@@ -55,11 +55,7 @@ class CustomSalesInvoice(SalesInvoice):
 
 	def make_pos_gl_entries(self, gl_entries):
 		"""
-		Override to add party information for wallet payment accounts.
-
-		The standard ERPNext implementation doesn't set party_type/party for
-		payment mode accounts, which causes validation errors for Receivable
-		accounts (like wallet accounts).
+		Rebase ERPNext 16 POS GL logic and add wallet-specific party info.
 		"""
 		if cint(self.is_pos):
 			skip_change_gl_entries = not _get_post_change_gl_entries_setting()
@@ -68,9 +64,12 @@ class CustomSalesInvoice(SalesInvoice):
 				if skip_change_gl_entries and payment_mode.account == self.account_for_change_amount:
 					payment_mode.base_amount -= flt(self.change_amount)
 
-				if payment_mode.amount:
+				against_voucher = self.name
+				if self.is_return and self.return_against and not self.update_outstanding_for_self:
+					against_voucher = self.return_against
+
+				if payment_mode.base_amount:
 					# POS, make payment entries
-					# Credit entry to debit_to (customer receivable)
 					gl_entries.append(
 						self.get_gl_dict(
 							{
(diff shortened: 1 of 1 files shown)


my-dev-jour:
Staging validation on ERPNext 16.11.0 showed the immediate root cause for the unpaid submitted invoice path was payment rows being overwritten during draft creation in `pos_next/api/invoices.py`, not the GL override alone.

The confirmed fix is now in PR #215, which preserves cashier-entered payment rows across `set_missing_values()` and was validated on staging with a real POS flow.

Closing or deprioritizing this PR in favor of #215 would avoid parallel review on the wrong root-cause path.

engahmed1190 (changes requested):
## Code Review — REQUEST CHANGES ❌ (BLOCKER)

### Summary
Rebases `make_pos_gl_entries` override on ERPNext 16 core flow. Adds `against_voucher` guard, `transaction_currency` fields, and removes the v15/v16 compatibility shim.

### Test Results
No tests included.

### CRITICAL BUG: v15 Breakage

**The PR removes the `hasattr(self, 'get_gle_for_change_amount')` compatibility check** and unconditionally calls `self.get_gle_for_change_amount()` (line 122).

The installed ERPNext on this bench is **v15**. In ERPNext v15, this method is called `make_gle_for_change_amount` (with a different signature — it takes `gl_entries` as a parameter). `get_gle_for_change_amount` does **not exist** in v15.

**This will cause an `AttributeError` at runtime** whenever a POS invoice with change amount is submitted. This is a hard crash in the invoice submission path.

The `hasattr` guard that was removed existed precisely for this reason — it allowed POSNext to work on both v15 and v16.

### Other Changes (Correct)

The non-breaking changes in this PR are good:
- **`against_voucher` pre-computation**: Cleaner than the inline ternary. Correct for returns.
- **`payment_mode.base_amount` gating**: Correct — matches ERPNext 16 core and handles multi-currency properly
- **`credit_in_transaction_currency` / `debit_in_transaction_currency`**: Correct for multi-currency GL entries. Harmless on v15 (`get_gl_dict` ignores unknown keys)
- **Docstring update**: Appropriate

### Required Fix
**Either:**
1. Restore the `hasattr` guard:
   ```python
   if hasattr(self, 'get_gle_for_change_amount'):
       gl_entries.extend(self.get_gle_for_change_amount())
   else:
       self.make_gle_for_change_amount(gl_entries)
   ```
2. OR defer this entire PR until the ERPNext 16 upgrade is confirmed and executed

### Additional Notes
- PR #215 description says it "supersedes the earlier GL-focused hypothesis from PR #209" — the payment-wiping bug was ultimately fixed at the `set_missing_values` level, not the GL level
- The `_get_post_change_gl_entries_setting()` helper is preserved (good)

### Verdict
The v15 compatibility removal is a runtime crash blocker. The other changes are correct and should be preserved, but the `hasattr` guard must be restored unless this project has committed to dropping v15 support.

PR #216, opened 2026-03-27
fix: make partial payments transactional and use core currency setup

### LINKED ISSUE #197: [Bug]: Partial payment API is not atomic and can leave Payment Entries after failure
### Bug Description
The partial payment flow commits `Payment Entry` documents inside helper-level code, while the outer API still behaves as if the overall operation can be rolled back on failure.

If one payment entry is committed and a later step fails, the request can return failure while one or more `Payment Entry` documents already exist.

### Expected Behavior
Partial payment submission should be atomic:
- either all related payment entries are created successfully
- or none are persisted

### Actual Behavior
The flow uses internal commits and then relies on outer rollback / compensating cancellation, which is not equivalent to real transactional rollback.

### Steps to Reproduce
1. Start a partial payment flow that creates more than one payment-related record
2. Trigger a failure after at least one `Payment Entry` has been created
3. Inspect the accounting documents after the API returns an error

### Impact
- accounting state can remain partially persisted after a failed request
- operators can see failed payment attempts that still created payment documents
- cleanup depends on best-effort cancellation recovery, which can also fail

### Suggested Fix
- remove helper-level commits from the partial payment path
- keep transaction control only at the orchestration layer
- if true atomicity is not supported, document the operation as best-effort instead of rollback-safe

### Notes
This appears separate from incorrect paid/outstanding amount issues. The core problem here is transaction safety.

### LINKED ISSUE #199: [Bug]: Partial payment Payment Entry creation uses wrong account currencies in multi-currency setups
### Bug Description
The partial payment flow builds `Payment Entry` documents manually and sets both `paid_from_account_currency` and `paid_to_account_currency` from the invoice currency.

That diverges from ERPNext core `Payment Entry` logic, where party-account currency and bank/cash-account currency are resolved separately.

### Expected Behavior
Partial payment `Payment Entry` creation should follow ERPNext core currency/account setup for:
- party account currency
- bank/cash account currency
- exchange-rate handling
- reference allocation

### Actual Behavior
The custom partial payment path hardcodes account currencies from the invoice currency.

### Steps to Reproduce
1. Use POSNext partial payments in a multi-currency environment
2. Create a sale where invoice currency differs from one or more linked accounts
3. Add a partial payment through the POSNext flow
4. Inspect the created `Payment Entry` values and resulting paid/outstanding behavior

### Impact
- incorrect paid/outstanding amounts in multi-currency sites
- wrong exchange-rate or account-currency behavior
- divergence from ERPNext core accounting logic

### Suggested Fix
- use ERPNext core `get_payment_entry(...)` flow where possible
- or mirror ERPNext's account-currency resolution logic exactly instead of deriving both sides from invoice currency

### Notes
This is separate from the partial-payment atomicity issue. The problem here is incorrect multi-currency setup even when the transaction itself succeeds.

my-dev-jour:
## Summary
- replace manual Payment Entry construction with ERPNext's `get_payment_entry(...)`
- remove helper-level commits from partial payment creation
- rollback the full payment batch to a savepoint if any payment in the request fails

## Why
`add_payment_to_partial_invoice()` currently claims rollback-safe behavior, but `create_payment_entry()` commits each submitted `Payment Entry` immediately. If a later payment in the same batch fails, the API returns an error after already creating accounting side effects and then tries to clean up with compensating cancels.

The current code also hardcodes both `paid_from_account_currency` and `paid_to_account_currency` from `invoice.currency`, which diverges from ERPNext's standard `Payment Entry` setup and is wrong for multi-currency sites.

## What changed
- use ERPNext core `get_payment_entry(...)` so account currencies and references come from the standard path
- keep POSNext-specific fields like `mode_of_payment`, `reference_no`, `remarks`, and explicit payment account selection
- wrap the batch in a database savepoint and roll back to that savepoint on failure
- remove the misleading compensating-cancel flow and helper-level `frappe.db.commit()`

## Validation
- reproduced the non-atomic behavior on ERPNext 16.11.0 staging before the fix: a failed batch still created a submitted `Payment Entry` and then cancelled it
- applied this patch on the same staging bench and re-tested a failed mixed-valid/invalid payment batch
- after the fix, no `Payment Entry Reference` rows remained for the invoice after the failed request

## Notes
- the rollback behavior in `#197` was runtime-validated on staging
- the multi-currency part of `#199` is a correctness fix aligned with ERPNext core, but staging is single-currency so that part was not runtime-proven there
- closes #197
- closes #199

### DIFF
--- a/pos_next/api/partial_payments.py
+++ b/pos_next/api/partial_payments.py
@@ -427,30 +427,15 @@ def create_payment_entry(
     if not frappe.db.exists("Mode of Payment", mode_of_payment):
         frappe.throw(_("Mode of Payment {0} does not exist").format(mode_of_payment))
 
-    # Create Payment Entry using ORM
-    pe = frappe.new_doc("Payment Entry")
-    pe.payment_type = "Receive"
-    pe.posting_date = posting_date
-    pe.party_type = "Customer"
-    pe.party = invoice.customer
-    pe.company = invoice.company
-    pe.mode_of_payment = mode_of_payment
+    # Save and submit with proper error handling
+    try:
+        from erpnext.accounts.doctype.payment_entry.payment_entry import get_payment_entry
+        from erpnext.accounts.doctype.sales_invoice.sales_invoice import get_bank_cash_account
 
-    # Set accounts
-    pe.paid_from = invoice.debit_to  # Customer receivable account
-
-    if payment_account:
-        # Validate provided account
-        if not frappe.db.exists("Account", payment_account):
-            frappe.throw(_("Payment account {0} does not exist").format(payment_account))
-        pe.paid_to = payment_account
-    else:
-        # Get account from Mode of Payment using ERPNext standard method
-        try:
-            from erpnext.accounts.doctype.sales_invoice.sales_invoice import (
-                get_bank_cash_account,
-            )
-
+        if payment_account:
+            if not frappe.db.exists("Account", payment_account):
+                frappe.throw(_("Payment account {0} does not exist").format(payment_account))
+        else:
             account_info = get_bank_cash_account(mode_of_payment, invoice.company)
             if not account_info or not account_info.get("account"):
                 frappe.throw(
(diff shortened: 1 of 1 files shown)


engahmed1190 (approved):
## Code Review — APPROVE ✅

### Summary
Replaces fragile manual Payment Entry construction with ERPNext's `get_payment_entry()` factory and replaces the broken compensating-cancel rollback with database savepoints.

### Test Results
No test file included. The logic change is straightforward and was validated on staging per the PR description.

### Correctness
- **`get_payment_entry()` replacement**: Correct. The old manual construction hardcoded `paid_from_account_currency` and `paid_to_account_currency` from `invoice.currency`, which is wrong for multi-currency sites. `get_payment_entry()` handles exchange rates, party accounts, and reference allocation correctly via ERPNext core.
- **Savepoint pattern**: Significant improvement. The old compensating-cancel approach had a serious problem: cancelling a submitted Payment Entry creates cancellation GL entries, so even after "rollback" the ledger contained submit + cancel noise. With `frappe.db.savepoint()` + `rollback(save_point=...)`, all database changes are undone at the SQL level — no phantom GL entries.
- **Removed `frappe.db.commit()`**: Correct and necessary for savepoints to work. The old per-payment commit made transactional rollback impossible.
- **Removed explicit `pe.validate()`**: Fine — `pe.insert()` already calls `validate()` internally, and `pe.submit()` also runs before_submit hooks.

### Security
No new concerns. `ignore_permissions` flag is kept with same justification. Length limits on `reference_no` (140) and `remarks` (500) are preserved.

1. Add `frappe.db.release_savepoint(batch_savepoint)` after the successful loop completion — currently relies on implicit release at transaction commit, which is fine but explicit is cleaner
2. Consider adding a test for the batch-with-failure scenario (e.g., second payment in a batch fails, verify first payment is also rolled back)
3. The `batch_savepoint` name could theoretically collide if this function is called recursively — extremely unlikely in practice but a unique suffix would be safer

### Verdict
Good improvement to both correctness (core currency setup) and reliability (savepoint atomicity). Ready to merge.

engahmed1190 (approved):
Confirming approval — logic is correct, savepoint pattern is solid, no test failures. Safe to merge.

PR #218, opened 2026-03-28
fix: submit customer credit redemption payload from POS

### LINKED ISSUE #210: [Bug]: Customer Credit UI counts payment as settled but submit flow does not send redemption payload
### Bug Description
Customer credit is added in the POS UI as a special payment entry, and the dialog counts it toward `totalPaid`, so the cashier can complete the invoice as if it were fully paid.

However, the invoice submit payload strips the customer-credit metadata and sends only plain payment rows. The backend only performs actual customer-credit redemption when it receives `customer_credit_dict` together with `redeemed_customer_credit`.

I could not find the frontend sending those redemption fields in the normal submit flow.

### Why This Is Distinct From #192
Issue `#192` covers the first failure mode where POSNext tries to resolve a GL account for `mode_of_payment = "Customer Credit"` and errors before save.

This issue is the deeper frontend/backend contract bug:
- the UI treats customer credit as settled payment
- but the normal submit payload does not send the backend redemption payload
- so even if the account-lookup failure is bypassed, the intended redemption path still is not invoked correctly

### Expected Behavior
When customer credit is applied in the payment dialog, the submit request should invoke the backend customer-credit redemption path correctly.

That means either:
- the frontend sends `customer_credit_dict` and `redeemed_customer_credit`, or
- the backend translates the special payment entry into that redemption path before submit

### Actual Behavior
The frontend counts customer credit as payment, but the normal invoice submission payload only sends plain `payments` rows.

### Steps to Reproduce
1. Enable customer credit payment in POS Settings
2. Give a customer available credit from a return / prior credit source
3. Open a new sale for that customer
4. Apply `Customer Credit` in the payment dialog
5. Complete the invoice flow
6. Inspect the payload/behavior versus the backend redemption path

### Impact
- the cashier can believe the invoice is fully paid when customer credit has not actually been redeemed through the intended backend flow
- invoices can remain unpaid/outstanding despite UI payment completion expectations
- fixing only the account-lookup error in `#192` would still leave an incomplete customer-credit payment flow

### Suggested Fix
- send `customer_credit_dict` and `redeemed_customer_credit` from the frontend when customer credit is applied
- or convert the special customer-credit payment entry into the backend redemption payload before submit
- do not let the UI count customer credit as settled payment unless the backend redemption path is actually engaged

### Notes
This issue is about the frontend/backend contract for customer-credit redemption, not just the missing account error already reported in `#192`.

my-dev-jour:
## Summary

Fix customer-credit POS sales so the frontend submits the redemption payload the backend already expects.

Before this change:
- the payment dialog created `Customer Credit` entries with `is_customer_credit` and `credit_details`
- `POSSale.vue` stripped that metadata before submit
- `submit_invoice()` never received `customer_credit_dict` / `redeemed_customer_credit`
- pure customer-credit sales could not redeem credit correctly

This patch:
- preserves customer-credit metadata through the POS submit flow
- builds `customer_credit_dict` and `redeemed_customer_credit` in `useInvoice.js`
- excludes synthetic `Customer Credit` rows from normal invoice `payments`
- allows pure customer-credit POS sales to pass core POS payment-row validation only when POSNext has explicitly marked the invoice for redemption

## Validation

- reproduced the bug on staging (`frappe 16.12.0`, `erpnext 16.11.0`, `pos_next 1.15.0`)
- applied the fix on staging and reran the same bench-console sale flow
- confirmed the invoice submitted successfully
- confirmed customer-credit redemption created a linked Journal Entry
- confirmed the saved invoice finished with `outstanding_amount = 0`

## Issue

Closes #210

### DIFF
--- a/POS/src/composables/useInvoice.js
+++ b/POS/src/composables/useInvoice.js
@@ -820,6 +820,76 @@ export function useInvoice() {
 		}
 	}
 
+	function serializeInvoicePayments(rawPayments) {
+		return rawPayments
+			.filter((payment) => !payment?.is_customer_credit)
+			.map((payment) => ({
+				mode_of_payment: payment.mode_of_payment,
+				amount: payment.amount,
+				type: payment.type,
+			}))
+	}
+
+	function buildCustomerCreditPayload(rawPayments) {
+		const creditPayments = rawPayments.filter((payment) => payment?.is_customer_credit)
+
+		if (!creditPayments.length) {
+			return {
+				invoicePayments: serializeInvoicePayments(rawPayments),
+				redeemedCustomerCredit: 0,
+				customerCreditDict: [],
+			}
+		}
+
+		const creditSources = new Map()
+		for (const payment of creditPayments) {
+			for (const credit of payment.credit_details || []) {
+				if (!credit?.type || !credit?.credit_origin) continue
+				const key = `${credit.type}:${credit.credit_origin}`
+				if (!creditSources.has(key)) {
+					creditSources.set(key, credit)
+				}
+			}
+		}
+
+		const redeemedCustomerCredit = roundCurrency(
+			creditPayments.reduce((sum, payment) => sum + Number(payment.amount || 0), 0),
+		)
+
+		let remainingCreditToAllocate = redeemedCustomerCredit
+		const customerCreditDict = []
+
+		for (const credit of creditSources.values()) {
+			if (remainingCreditToAllocate <= 0) break
+
+			const availableCredit = roundCurrency(
+				Number(credit.available_credit ?? credit.total_credit ?? 0),
+			)
+			if (availableCredit <= 0) continue
+
+			const creditToRedeem = Math.min(availableCredit, remainingCreditToAllocate)
+			if (creditToRedeem <= 0) continue
+
+			customerCreditDict.push({
+				...credit,
+				credit_to_redeem: roundCurrency(creditToRedeem),
+			})
+			remainingCreditToAllocate = roundCurrency(
+				remainingCreditToAllocate - creditToRedeem,
+			)
+		}
+
+		if (remainingCreditToAllocate > 0.01) {
+			throw new Error("Unable to allocate the selected customer credit")
+		}
+
+		return {
+			invoicePayments: serializeInvoicePayments(rawPayments),
+			redeemedCustomerCredit,
+			customerCreditDict,
+		}
+	}
+
 	async function saveDraft(targetDoctype = "Sales Invoice") {
 		/**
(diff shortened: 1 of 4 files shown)


engahmed1190 (approved):
## Code Review — APPROVE (with suggestions) ✅

### Summary
Enables pure customer-credit POS sales by preserving credit metadata through the submit flow and adding a scoped `validate_pos_paid_amount` override.

### Test Results
No test file included. Validated on staging per PR description.

### Correctness

**Frontend (`buildCustomerCreditPayload`):**
- Well-structured: separates credit payments from regular payments, deduplicates credit sources via Map, allocates FIFO with tolerance check (0.01)
- Edge case handled: if `credit_details` is missing on an `is_customer_credit` payment, the allocation throws — correct, better to fail than silently lose credit
- `serializeInvoicePayments` correctly filters out `is_customer_credit` payments from the normal payment array

**Backend (`validate_pos_paid_amount`):**
- Correctly scoped: only activates when `pos_next_redeemed_customer_credit` flag is set AND zero payments AND POS invoice with positive grand total
- Falls through to `super().validate_pos_paid_amount()` for all other cases
- The flag is on `invoice_doc.flags` (transient Python attribute, not a DocField) — cannot be injected via Frappe's API or JSON payload

**`POSSale.vue` (`...p` spread):**
- Correctly preserves `is_customer_credit` and `credit_details` metadata that was previously stripped

### Security
The `flags` pattern is idiomatic Frappe and safe. The only code path that sets `pos_next_redeemed_customer_credit` is through `submit_invoice()` which is behind `@frappe.whitelist()`.

1. **Add a backend guard**: if `redeemed_customer_credit > 0` but `customer_credit_dict` is empty/falsy, don't set the flag. Currently, a crafted API call with `redeemed_customer_credit=100` and no `customer_credit_dict` would submit a zero-payment invoice that stays fully outstanding
2. **Add tests** for the `validate_pos_paid_amount` override — verify it bypasses with the flag and does NOT bypass without it
3. **Consider filtering the `...p` spread** in POSSale.vue to only known properties, to prevent unexpected properties leaking into the cart store

### Conflict Note
This PR modifies `submit_invoice` in `invoices.py` and adds `validate_pos_paid_amount` to `sales_invoice.py`. No conflict with the current `fix/payment-amount-wiped-by-erpnext` branch (which modifies `update_invoice`).

### Verdict
Well-scoped fix. The `flags` pattern is the right approach for a temporary validation bypass. Good to merge.

engahmed1190 (approved):
Confirming approval — customer credit payload fix is correct, validate_pos_paid_amount override is properly scoped via flags pattern. Safe to merge.

PR #219, opened 2026-03-28
fix: honor coupon apply_on in POS

my-dev-jour:
## Summary
- honor coupon `apply_on` when computing the discount amount in the POS coupon dialog
- use the same selected base amount when applying the coupon to the cart so percentage coupons are not recalculated against subtotal
- pass tax and grand total into the coupon dialog so `Grand Total` and `Net Total` coupons produce different results as intended

## Root cause
`CouponDialog.vue` always validated `min_amount` and calculated the coupon from `subtotal`, even when the coupon itself carried `apply_on = "Grand Total"` or `apply_on = "Net Total"`.

The stale frontend calculation was then repeated in `useInvoice.applyDiscount()`, which recalculated percentage discounts against `subtotal` again.

## Validation
- confirmed on the staging code path before patching: coupon dialog ignored `coupon.apply_on` and always used `subtotal`
- confirmed current upstream code still had the same behavior before patching
- deployed the patch to a staging environment and rebuilt POS assets
- local verification was limited to `git diff --check` in this workspace

### DIFF
--- a/POS/src/components/sale/CouponDialog.vue
+++ b/POS/src/components/sale/CouponDialog.vue
@@ -162,6 +162,14 @@ const props = defineProps({
 		required: true,
 		note: __("Cart subtotal BEFORE tax - used for discount calculations"),
 	},
+	taxAmount: {
+		type: Number,
+		default: 0,
+	},
+	grandTotal: {
+		type: Number,
+		default: 0,
+	},
 	items: Array,
 	posProfile: String,
 	customer: String,
@@ -257,6 +265,14 @@ function applyGiftCard(card) {
 	applyCoupon()
 }
 
+function getCouponBaseAmount(coupon) {
+	const grandTotal = Number.parseFloat(props.grandTotal || 0)
+	const taxAmount = Number.parseFloat(props.taxAmount || 0)
+	const netTotal = Math.max(grandTotal - taxAmount, 0)
+
+	return coupon.apply_on === "Grand Total" ? grandTotal : netTotal
+}
+
 async function applyCoupon() {
 	if (!couponCode.value.trim()) {
 		errorMessage.value = __("Please enter a coupon code")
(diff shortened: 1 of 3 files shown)


engahmed1190 (approved):
## Code Review — APPROVE ✅

### Summary
Frontend fix: CouponDialog.vue previously always calculated discounts against `subtotal` regardless of the coupon's `apply_on` setting. Now respects `Grand Total` vs `Net Total` coupon configurations.

### Test Results
No tests (frontend-only change). Validated via staging per PR description.

### Correctness
- `getCouponBaseAmount()` correctly branches on `coupon.apply_on === "Grand Total"`
- `base_amount` is propagated to `useInvoice.applyDiscount()` via the discount object, ensuring the same base is used for both dialog preview and actual discount application
- The `typeof discount.base_amount === "number"` check in `useInvoice.js` correctly handles `base_amount === 0` (doesn't fall through to subtotal)
- `Number.parseFloat(props.grandTotal || 0)` handles both `undefined` and `0` correctly since props have `default: 0`

### Code Quality
Clean. New props (`taxAmount`, `grandTotal`) with proper type and defaults. Good separation of `getCouponBaseAmount` as a focused helper.

1. **Consider using `props.subtotal` for the Net Total case** instead of computing `grandTotal - taxAmount`. The component already receives `subtotal` (described as "cart subtotal BEFORE tax"), which should be identical to net total. Using `subtotal` directly avoids introducing a potential rounding discrepancy:
   ```javascript
   return coupon.apply_on === "Grand Total" ? grandTotal : props.subtotal
   ```
   This would also eliminate the need for the `taxAmount` prop entirely.

2. The `calculateDiscountAmount` docstring in `useInvoice.js` says "Discounts are ALWAYS calculated on SUBTOTAL" which is no longer true. Worth updating if you touch that area.

3. The `base_amount` is computed at coupon-application time and stored in the discount object. If the cart changes after coupon application, `base_amount` becomes stale. This matches existing behavior (coupons must be re-applied after cart changes), so it's fine.

### Verdict
Clean, well-scoped fix. Good to merge.

engahmed1190 (approved):
Confirming approval — clean frontend fix, no test failures. Safe to merge.

PR #220, opened 2026-03-28
fix: scope wallet reservations to draft invoices

my-dev-jour:
## Summary
- only reserve wallet balance for draft POS invoices, not submitted invoices already reflected in GL
- scope pending wallet reservations to the same company as the wallet lookup
- add regression coverage for both wallet helper implementations

## Problem
Wallet balance lookup subtracted pending wallet payments using all POS invoices with `docstatus in [0, 1]` and without company scoping.

That meant a submitted POS invoice could reduce available wallet balance a second time even though its effect was already in GL, and reservations could leak across companies.

## Staging Reproduction
Confirmed on a staging environment (ERPNext `16.11.0` / POSNext `1.15.0`):
- mocked wallet GL balance to `300`
- a submitted partial POS invoice with outstanding `100` and a wallet-flagged payment reduced available balance from `300` to `200`
- this proved submitted invoices were being subtracted again as pending reservations

## Fix
- change pending reservation queries to draft invoices only (`docstatus = 0`)
- pass company through wallet balance and available balance calculations
- keep `exclude_invoice` behavior intact

## Verification
- staging reproduction confirmed the original bug

### DIFF
--- a/pos_next/api/test_wallet.py
+++ b/pos_next/api/test_wallet.py
@@ -1,0 +1,89 @@
+# Copyright (c) 2025, BrainWise and contributors
+# For license information, please see license.txt
+
+import unittest
+from types import SimpleNamespace
+from unittest.mock import Mock, patch
+
+from pos_next.api import wallet as api_wallet
+from pos_next.pos_next.doctype.wallet import wallet as wallet_doctype
+
+
+class TestWalletAPI(unittest.TestCase):
+	@patch("pos_next.api.wallet.frappe.db")
+	@patch("pos_next.api.wallet.frappe.get_all")
+	def test_get_pending_wallet_payments_only_counts_draft_wallet_payments_in_same_company(
+		self,
+		mock_get_all,
+		mock_db,
+	):
+		mock_get_all.side_effect = [
+			[
+				SimpleNamespace(name="DRAFT-SAME"),
+				SimpleNamespace(name="EXCLUDED-DRAFT"),
+			],
+			[
+				SimpleNamespace(mode_of_payment="Wallet", amount=40),
+				SimpleNamespace(mode_of_payment="Cash", amount=10),
+			],
+		]
+		mock_db.get_value.side_effect = lambda doctype, name, field: 1 if name == "Wallet" else 0
+
+		result = api_wallet.get_pending_wallet_payments(
+			"Guest",
+			exclude_invoice="EXCLUDED-DRAFT",
+			company="Sonex",
+		)
+
+		self.assertEqual(result, 40)
+		self.assertEqual(mock_get_all.call_count, 2)
+		self.assertEqual(
+			mock_get_all.call_args_list[0].kwargs["filters"],
+			{
+				"customer": "Guest",
+				"docstatus": 0,
+				"outstanding_amount": [">", 0],
+				"is_pos": 1,
+				"company": "Sonex",
+			},
+		)
+
+	@patch("pos_next.pos_next.doctype.wallet.wallet.get_pending_wallet_payments")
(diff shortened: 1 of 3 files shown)


engahmed1190 (changes requested):
## Code Review — REQUEST CHANGES ❌

### Summary
Changes wallet pending reservation queries from `docstatus in [0, 1]` to `docstatus = 0` (draft only) and adds company scoping. Applied in both `pos_next/api/wallet.py` and `pos_next/pos_next/doctype/wallet/wallet.py`.

### Test Results
**2/3 tests pass**, 1 fails. `test_get_pending_wallet_payments_only_counts_draft_wallet_payments_in_same_company` fails — the `frappe.db.get_value` mock doesn't properly intercept the Werkzeug LocalProxy, so the real `frappe.db.get_value` is called against the database.

### Correctness

**Core fix is correct.** The old code double-counted wallet reservations: submitted invoices are already reflected in GL via `get_balance_on()`, so counting them again as "pending" artificially deflated available balance. Draft-only (`docstatus = 0`) correctly captures invoices that reserve wallet balance but haven't posted to GL.

**Company scoping is correct.** Without it, wallet reservations in Company A reduced available balance for Company B. The optional `company` parameter preserves backward compatibility.

**Edge case — submission window:** Brief moment between `docstatus` changing to 1 and GL entries being committed. This is within a single database transaction, so concurrent requests see either full-draft or full-submitted state — safe.

### Issue: Code Duplication (NEEDS DISCUSSION)

There are **TWO near-identical implementations** of `get_pending_wallet_payments` and `get_customer_wallet_balance`:
- `pos_next/api/wallet.py` (lines 220-258)
- `pos_next/pos_next/doctype/wallet/wallet.py` (lines 136-182)

Both were patched identically in this PR, which demonstrates the problem — a future developer may fix a bug in one copy but not the other. This is a maintenance hazard.

**Recommendation:** One implementation should import from the other, or both should delegate to a shared utility. This doesn't need to block this PR, but should be tracked and addressed.

### Required Fixes
1. **Fix the failing test** — mock `frappe.db` at `frappe.local.db` level or use the `frappe.qb` pattern (which works correctly, as demonstrated by PR #221's passing tests)
2. **Add a code comment** noting the duplication and linking to a follow-up task, OR consolidate the implementations in this PR

### Additional Suggestions (non-blocking)
1. **Missing test**: No test for `exclude_invoice` logic — verify that an invoice matching `exclude_invoice` is skipped from the total
2. The `Wallet.get_available_balance` change (line 57) correctly uses `self.company` — good

### Verdict
The logic fix is correct and well-reasoned. Fix the test failure and address the duplication concern (at minimum acknowledge it), then this is ready to merge.

my-dev-jour:
Addressed the requested follow-up in branch commit c4272cb. I kept the wallet logic fix intact and made two reviewer-driven changes: 1) the failing test no longer mocks  through the LocalProxy path; the API path now uses a small  helper so the draft/company reservation test can patch a stable function instead, and 2) I removed the duplicated reservation implementation in  by delegating both doctype helpers to the API implementation with explicit compatibility comments. The wrapper tests now cover that delegation path directly.

my-dev-jour:
Addressed the requested follow-up in branch commit c4272cb.

I kept the wallet logic fix intact and made two reviewer-driven changes:

1. The failing test no longer mocks `frappe.db.get_value` through the LocalProxy path.
   - the API path now uses a small `_is_wallet_payment_mode()` helper
   - the draft/company reservation test now patches that stable helper instead

2. I removed the duplicated reservation implementation in `pos_next/pos_next/doctype/wallet/wallet.py`.
   - both doctype helpers now delegate to the API implementation
   - I left explicit compatibility comments in the wrapper layer
   - the wrapper tests now cover that delegation path directly

PR #222, opened 2026-03-29
feat: auto-select quantity input when UOM dialog opens

NotAbdelrahmanelsayed:
## Summary

- Add `ref="quantityInput"` to the quantity `<input>` in `ItemSelectionDialog`
- Import `nextTick` from Vue
- When the dialog opens in `uom` mode, automatically focus **and select** the quantity field so users can immediately type a new quantity without manually clearing the default value

## Behaviour Before / After

| Before | After |
|--------|-------|
| Dialog opens, cursor is not in the quantity field | Dialog opens, quantity field is focused and its value is selected — type to overwrite instantly |

## Testing

1. Open POS and add an item that has multiple UOMs configured.
2. Click the item to open the UOM selection dialog.
3. Verify the quantity input is focused and the default value (`1`) is selected.
4. Type a new quantity — it should replace the selected value immediately.

### DIFF
--- a/POS/src/components/sale/ItemSelectionDialog.vue
+++ b/POS/src/components/sale/ItemSelectionDialog.vue
@@ -118,6 +118,7 @@
 							</button>
 							<div class="flex-1 h-full flex items-center justify-center px-3">
 								<input
+									ref="quantityInput"
 									v-model.number="quantity"
 									type="number"
 									min="1"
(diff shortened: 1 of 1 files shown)


NotAbdelrahmanelsayed:
https://github.com/user-attachments/assets/7b55a272-79e5-4f76-87c6-9a9d160e8069

PR #231, opened 2026-04-04
feat: keyboard-only payment completion from desktop dialog

NotAbdelrahmanelsayed:
## Summary

- Replaces the first quick-amount button in the desktop payment grid (`div.hidden.lg:block > div.grid.grid-cols-4`) with a currency input field
- Input is pre-filled with the grand total and auto-focused/selected when the dialog opens — cashier can press **Enter** immediately to complete payment, no mouse needed
- Typing a different amount before pressing Enter applies that amount instead; the remaining 3 quick-amount buttons are unchanged
- Extends the mobile custom-amount input and numpad Enter key with the same auto-complete behaviour

## Demo

**Before:**

https://github.com/user-attachments/assets/56dc75d3-d9af-4b39-b550-42b514694f42

**After:**

https://github.com/user-attachments/assets/e4edd5c4-1cde-41a2-83bf-4f4134df2b35

## How it works

1. Dialog opens → `desktopAmountInput` is pre-filled with grand total and the input is focused/selected
2. Cashier presses Enter → `handleDesktopAmountEnter` calls `addCustomPayment`, then `completePayment` if fully paid
3. `@keydown.enter.stop.prevent` stops the event from bubbling to the numpad's global `window` listener (prevents double-counting)

### DIFF
--- a/POS/src/components/sale/InvoiceCart.vue
+++ b/POS/src/components/sale/InvoiceCart.vue
@@ -192,6 +192,7 @@
 
 							<!-- Native Input for Instant Search -->
 							<input
+								ref="customerSearchInputRef"
 								id="cart-customer-search"
 								name="cart-customer-search"
 								:value="customerSearch"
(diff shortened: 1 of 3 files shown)


NotAbdelrahmanelsayed:
Closing in favour of a clean single-commit PR.

PR #239, opened 2026-04-15
feat: select first search result on Enter key press

NotAbdelrahmanelsayed:
## Summary

- When the cashier types a partial item name in the search box and presses **Enter**, the **first visible search result** is added to the cart immediately
- If no search results are present (e.g. typing a barcode), the existing exact barcode lookup runs as before — no regression for scanner workflows
- Variant/UOM/batch dialogs still appear when needed (respects auto-add mode)

## Behaviour change

| Before | After |
|--------|-------|
| Enter → exact barcode lookup → "Item Not Found" for partial text | Enter → first search result added to cart |
| Barcode scan → works as before | Barcode scan → works as before (fallback path) |

## Files changed

- `POS/src/composables/useSearchInput.js` — `handleKeyDown`: checks `itemStore.filteredItems[0]` before falling back to barcode lookup

## Test plan

- [ ] Type a partial item name → results appear → press Enter → first result added to cart
- [ ] Type an exact barcode → press Enter → barcode lookup fires normally
- [ ] Type something with no matches → press Enter → "Item Not Found" warning
- [ ] With auto-add ON: Enter adds without prompts for simple items
- [ ] With auto-add OFF: Enter opens variant/UOM dialog if applicable


NotAbdelrahmanelsayed:
Recreated as #368 since GitHub won't let a bot-closed PR be reopened via the API. Same branch, no code changes — following up there.

PR #249, opened 2026-05-07
fix: MRP and MSP in List View POS , Pin Code in the new customer and Rate Editable in the POS for non stock item

zaidd29:


engahmed1190:
Please add screenshoot or video

matiyas-develop:
@zaidd29 please add screenshots or videos

MohamedAliSmk:
### Critical — Should Fix Before Merge

#### 1. Bug in `EditItemDialog.vue` — Wrong Ref Access

**File:** `POS/src/components/sale/EditItemDialog.vue` (line ~696)

```javascript
if ((settingsStore.allowUserToEditRate || !localItem.is_stock_item) && isRateManuallyEdited) {
```

`localItem` is a `ref`, so `localItem.is_stock_item` is always `undefined`, making `!localItem.is_stock_item` always `true`. This means max-discount validation runs for **every** manual edit regardless of POS settings.

**Fix:** Use `!localItem.value?.is_stock_item` instead.

---

#### 2. Server Trusts Client `is_stock_item` for Rate-Edit Bypass

**File:** `pos_next/api/invoices.py` (lines ~143–150)

```python
is_stock_item = cint(item.get("is_stock_item") or 0)

# Check if rate editing is allowed for stock items only
if is_stock_item and not cint(pos_settings.get(FIELD_ALLOW_USER_TO_EDIT_RATE)):
```

Stock validation elsewhere looks up `is_stock_item` from the Item master; this path does not. A malicious client could send `is_stock_item: 0` on a stock item to bypass rate-edit restrictions.

**Fix:** Look up `is_stock_item` from the Item doctype, consistent with stock validation elsewhere in `invoices.py`.

---

#### 3. Accidental Regression in `get_items_bulk` — `selling == 1` Filter Removed

On `develop`, the bulk price query includes:

```python
.where(ItemPrice.selling == 1)
```

This branch removes that filter. That can pull buying prices or wrong price rows into offline/bulk item loads.

**Fix:** Restore `.where(ItemPrice.selling == 1)` in `get_items_bulk`.

---

### Medium Priority

#### 4. Inconsistent UI — Grid vs List View

- **Grid view** (line ~386 in `ItemsSelector.vue`) still shows a single **Rate** column.
- **List view** shows **MRP** + **MSP**.

Cashiers may see different price information depending on view mode.

**Suggestion:** Add MRP/MSP to grid view or document the intentional difference.

---

#### 5. Pincode Required in UI, Optional in API

- Frontend enforces pincode as required.
- Backend docstring says `custom_pincode` is optional and accepts empty string.

API callers can still create customers without pincode, bypassing the UI rule.

**Suggestion:** Add server-side validation if pincode is a business requirement.

---

#### 6. `custom_pincode` / `mrp_rate` Not Defined in This Repo

No fixture or custom-field definition exists in the repo for:

- `Customer.custom_pincode`
- `Item Price.mrp_rate`

These fields must already exist on the target ERPNext site, or the branch will fail at runtime.

**Suggestion:** Include custom field definitions in the PR or document the site dependency.

---

#### 7. Edit Existing Customers Without Pincode

Edit mode now requires pincode. Existing customers with no `custom_pincode` cannot be saved until one is entered.

**Suggestion:** Confirm this is intended business behavior.

---

### Low Priority / Nits

#### 8. Indentation in `CreateCustomerDialog.vue`

The pincode block is mis-indented compared to surrounding form fields.

---

#### 9. Unrelated `useQuickAmounts.js` Change

This change is not mentioned in the commit message and alters payment UX behavior. Consider:

- Reverting if unintentional, or
- Splitting into a separate commit/PR

The new step-based logic may produce less useful suggestions (e.g. ₹103 → ₹105, ₹110, ₹115 instead of rounding to ₹110, ₹150).

---

#### 10. Non-Stock Items Bypass Pricing-Rule Lock

Frontend allows rate edits on non-stock items even when promotional offers (pricing rules) are applied. Confirm this is intended with product/business stakeholders.

---

#### 11. Docstring Drift

`validate_manual_rate_edit` docstring still states that rate editing must be enabled in POS Settings, but non-stock items are now exempt. Update the docstring to reflect the new behavior.

PR #273, opened 2026-05-23
Fix/pos modal mouse focus

lubshad:
## Summary

Fix POS modal fields not focusing on mouse/tap.

## Changes

- Added a local `POSDialog` wrapper around `frappe-ui` `Dialog`.
- Stops modal content pointer events from reaching the overlay.
- Keeps input focus working while preventing unwanted drag/text selection in modal UI.


MohamedAliSmk:
@lubshad 
thanks for your efforts,
could you please review changes as recording video, to show in how can you fix POS modal fields not focusing on mouse/tap.
thanks.

PR #275, opened 2026-05-25
Feat/eod shift report print

MohamedAliSmk:
## Summary
- Fixes EOD / POS Closing Shift print: address lookup, money formatting (`fmt_money` + stored `difference`), thermal layout, and qty/rate display.
- Empty Items / Taxes / Payments no longer look like a clean close — section headers stay and show explicit empty lines.
- Early-return in `get_items_sold` when there are no parent targets; regression tests for empty transaction / empty parent-target cases.
- Button label translations for ar / id / pt-BR (report body still English — follow-up).

## Test plan
- [ ] Close a shift with items, taxes, payments → totals and Short/Over look right
- [ ] Close / print with empty items, taxes, or payments → headers + "No … rows found"
- [ ] `bench run-tests --app pos_next --module pos_next.pos_next.utils.tests.test_pos_closing_print`


pos_next/pos_next/utils/pos_closing_print.py
```diff
@@ -87,6 +87,9 @@ def _fetch_items_for_targets(parent_targets: set[tuple[str, str]]) -> list[dict]
 def get_items_sold(doc) -> list[dict]:
 	closing_doc = _as_closing_doc(doc)
 	parent_targets = _collect_parent_targets(closing_doc.get("pos_transactions"))
+	if not parent_targets:
```
engahmed1190, line 87: **This guard is a no-op — `_fetch_items_for_targets` already has it.**

`_fetch_items_for_targets` opens with the identical `if not parent_targets: return []` (line 58-59 of this file), so behaviour is unchanged either way. One of the two guards is dead code.

Pick one: keep it in `_fetch_items_for_targets` (closest to the query it protects) and revert this hunk, or move it up here and delete the inner one. Not both.

pos_next/pos_next/utils/tests/test_pos_closing_print.py
```diff
@@ -75,3 +75,11 @@ def test_get_items_sold_returns_empty_when_no_transactions(self, mock_fetch):
...
 		self.assertEqual(result, [])
 		mock_fetch.assert_not_called()
+
+	@patch("pos_next.pos_next.utils.pos_closing_print._fetch_items_for_targets")
+	def test_get_items_sold_returns_empty_when_no_parent_targets(self, mock_fetch):
+		doc = {"pos_transactions": [{"customer": "CUST-001"}]}
+		result = get_items_sold(doc)
+
+		self.assertEqual(result, [])
+		mock_fetch.assert_not_called()
```
engahmed1190: **This test asserts an implementation detail, not behaviour.**

`mock_fetch.assert_not_called()` locks in *which* of the two duplicate guards fires. Since `_fetch_items_for_targets` already returns `[]` for an empty set, `get_items_sold` returns `[]` either way — this test would fail on a functionally identical refactor while the observable behaviour is unchanged.

If the redundant guard in `get_items_sold` goes (see the other comment), this test goes with it. The `assertEqual(result, [])` half is worth keeping; the mock assertion is not.

pos_next/pos_next/print_format/pos_next_eod_report/pos_next_eod_report.json
```diff
@@ -8,7 +8,7 @@
...
...ed.v) }}\n        </span>\n    </div>\n\n    <div class=\"eod-row grand\">\n        <span class=\"eod-label\">Total Counted</span>\n        <span class=\"eod-value\">\n            {{ \"{:,.2f}\".format(total_counted.v) }}\n        </span>\n    </div>\n\n    <div class=\"eod-row grand\">\n\n        <span class=\"eod-label\">\n            Short / Over\n        </span>\n\n        <span class=\"eod-value\">\n\n            {{ \"{:,.2f}\".format(short_over) }}\n\n            {% if short_over < 0 %}\n                <span class=\"eod-tag short\">\n                    SHORT\n                </span>\n            {% elif short_over > 0 %}\n                <span class=\"eod-tag over\">\n                    OVER\n                </span>\n            {% endif %}\n\n        </span>\n\n    </div>\n\n    <!-- =========================================\n         FOOTER\n    ========================================== -->\n\n    <div class=\"eod-footer\">\n\n        Printed:\n        {{ frappe.utils.format_datetime(frappe.utils.now_datetime()) }}\n\n        <br>\n\n        by {{ frappe.session.user }}\n\n        <div class=\"eod-end\">\n            *** END ***\n        </div>\n\n    </div>\n\n</div>",
```
engahmed1190: Grouping the template findings here, since the whole rewrite is one line.

---

**1. Recomputing `difference` instead of using the stored field — Medium**

`pos_closing_shift.py:73` already computes and stores `d.difference = flt(closing, precision) - flt(expected, precision)` — precision-rounded. This template ignores it and recomputes `p.closing_amount - p.expected_amount` raw, then derives Short/Over from its own `namespace()` accumulators. That is two sources of truth for the same money figure: the printed receipt can disagree with the form view and with `pos_closing_shift.py:290`. The previous template used `payment.difference`, so this is a regression.

```jinja
{% set diff = p.difference or 0 %}
{% set short_over = doc.payment_reconciliation | sum(attribute='difference') %}
```

That also deletes the whole 6-line `namespace()` accumulator block.

---

**2. No currency anywhere on a financial report — Medium**

Every figure is hardcoded `"{:,.2f}".format(...)`: no symbol, no company currency precision, comma grouping regardless of locale. `POS Closing Shift` has no `currency` field, so it is the company default. `fmt_money` *is* whitelisted in the Jinja sandbox (checked frappe's `VALID_UTILS`), so this works:

```jinja
{{ frappe.utils.fmt_money(doc.grand_total, currency=company_doc.default_currency) }}
```

---

**3. `{:g}` sends quantities to scientific notation — Low**

`"{:g}".format(1000000.0)` renders `1e+06`; `1234567.0` renders `1.23457e+06`. Affects the Items Sold qty and the tax rate. Unlikely, free to fix.

---

**4. `get_doc('Address', ...).get_display()` breaks under restricted Jinja — Low**

Default sites are fine: `disable_render_safe_exec` defaults to `1`, so unrestricted globals give you the real `frappe.get_doc`. But on a site with render-safe-exec enabled, `frappe.get_doc` is `get_doc_as_dict` and returns a plain dict — `.get_display()` is then undefined and the entire print raises. It also loads a full Address doc for one string.

```jinja
{{ frappe.db.get_value('Address', profile_doc.company_address, 'address_display') }}
```

Works in both modes. Note `get_address_display` is *not* sandbox-whitelisted, so don't reach for that one.

---

**5. Grey fills on an 80mm thermal printer — Low**

`.eod-section-title` `#e5e5e5`, `.eod-row.grand` `#e9e9e9`, and the tags `#d9d9d9`/`#efefef`, all forced with `print-color-adjust: exact`. Thermal heads dither grey as halftone — slower and darker — and SHORT (`#d9d9d9`) vs OVER (`#efefef`) come out indistinguishable on paper. The text labels carry the meaning so it is not an accessibility failure, but the fills buy nothing; the previous template used dashed rules. Separately, `body { padding-bottom: 40px }` burns roughly 4mm of paper on every print.

engahmed1190 (commented):
## Full review — coding, quality, performance, security, standards, coverage

**Verdict:** ship-able after two changes — drop one of the two duplicate empty guards, and use the stored `difference` field instead of recomputing it. The rest is polish.

Real scope is much bigger than `+15/-1` suggests: this is a 6.5 KB Jinja print-format rewrite compressed into a single JSON string, plus a 3-line Python guard, plus 3 translation rows. Inline comments carry the specifics.

### What's good
- **Real bug fixed:** the old template read `frappe.db.get_value('POS Profile', ..., 'address')`. POS Profile has no `address` field (verified — only `company` and `company_address`), so the address block *never rendered*. The new `company_address` is correct. Genuine win.
- `format_datetime` over a raw datetime repr, and `user_doc.full_name or doc.user` over the raw user id. Both right.
- Translations complete: `"Print EOD Report"` added to all three CSVs, and the source string exists at both call sites (`pos_closing_shift.js:32`, `ShiftClosingDialog.vue:965`). No locale missed.

### Performance
Fine, slightly better than before. `get_items_sold` is one aggregated `Sales Invoice Item` query — no N+1. The template adds 3 `get_cached_doc` + 1 `get_doc`, all O(1), and no query inside any loop. Only note is the full `Address` doc load (inline).

### Security
**Nothing new introduced** — no SQL string building, no permission bypass, no new sink.

Pre-existing and worth knowing: frappe's `FrappeSandboxedEnvironment` runs with `autoescape=False` (`frappe/utils/jinja.py`). Every `{{ }}` here — item names, mode of payment, account head, company name — is raw HTML. An item named `<style>*{display:none}</style>` blanks the receipt. This applied equally to the old template. Close it with `| e` on the free-text fields (not on `get_display()`, which is intentionally HTML).

### Coverage
**Inverted.** The one added test covers a redundant no-op line. The substantive change — a rewrite of every money figure on a financial receipt — has zero coverage, and is untestable as written because the HTML is buried in a JSON string.
- Minimum: render the format against a fixture closing shift, assert the grand total and the Short/Over string appear.
- Bigger win: extract the template to a sibling `.html` file so it is diffable, lintable and reviewable. The single-line 6.5 KB JSON string is why this review had to be done by eye.

### Standards
- The report body is hardcoded English with `default_print_language: 'en'`, while this PR adds ar/id/pt-br for *the button that opens it*. An Arabic cashier gets an Arabic button and an English LTR receipt. Pre-existing, but a rewrite was the moment to wrap labels in `{{ _("...") }}` and set `direction` from `frappe.local.lang`.
- Python side follows repo conventions (tabs, `from __future__ import annotations`, typed helpers, `frappe.qb`). Clean.
- PR description is empty, on a PR whose diff GitHub cannot render usefully. Worth a paragraph.

pos_next/pos_next/utils/pos_closing_print.py
engahmed1190: **Correction to my comment above — I was wrong, don't revert this hunk.**

I said this guard was a pure no-op and suggested reverting it. It is behaviourally a no-op, but it is not gratuitous: the *pre-existing* test `test_get_items_sold_returns_empty_when_no_transactions` asserts `mock_fetch.assert_not_called()`, and without this guard `_fetch_items_for_targets` is called with an empty set. **That test is red on `develop` today**, and this hunk is what makes it green. Verified by running it against both branches:

```
develop:  FAIL - AssertionError: Expected '_fetch_items_for_targets' to not have been
          been called. Called 1 times. Calls: [call(set())]
pr-275:   ok
```

So the hunk earns its place. What still stands is the narrower point: there are now two identical `if not parent_targets: return []` guards (here and at line 58-59), and only one should survive. Keeping this one and deleting the inner one is the cleaner resolution, since it is the one the tests pin.

Apologies for the bad call — reverting as I first suggested would have re-broken a test.

pos_next/pos_next/utils/tests/test_pos_closing_print.py
engahmed1190: **Follow-up to my comment above.** I suggested this test could go along with the guard it covers. That was based on a wrong reading — see the correction on `pos_closing_print.py:90`: the guard is not revertible, because a pre-existing test depends on it (and is red on `develop` without it).

The narrower point stands on its own though: this test and the existing `test_get_items_sold_returns_empty_when_no_transactions` assert the same two things via near-identical inputs (empty list vs. a row with no invoice link). Worth keeping one, or merging them with `subTest`.

pos_next/pos_next/print_format/pos_next_eod_report/pos_next_eod_report.json
engahmed1190: Two more template findings, from a second pass — both about **silent** empty states on a report that is the shift's audit record.

---

**6. The Items Sold section vanishes entirely when there are no items**

The rewrite wraps the whole section in `{% if items %}`; the old template kept the header and printed `<div class="muted">No item rows found</div>` in an `{% else %}`. Combined with the new early return in `get_items_sold`, an empty result is now a normal, silent outcome — a closing shift whose item lookup yields nothing (e.g. `pos_transactions` rows with a blank `sales_invoice`) prints a receipt with no Items Sold section at all, indistinguishable from a genuinely item-less shift. Restore the `{% else %}` fallback so a data problem is visible on paper.

---

**7. Payments has no guard at all, and the summary then reads as "reconciled"**

Items and Taxes got `{% if %}` guards; Payments did not. The `<div class="eod-section-title">Payments</div>` is unconditional, so with an empty `payment_reconciliation` the report prints a dangling header followed by nothing, and then a Final Summary of `0.00 / 0.00 / 0.00` with no tag — which reads as *everything reconciled* rather than *no payment data*. This is the worst of the three empty states, because it is the one that produces a confidently wrong number rather than a blank.

Either guard it symmetrically with the other two sections, or (better for an audit document) keep all three headers unconditional and give each an explicit empty-state line.

engahmed1190 (commented):
Second pass — two additional findings, plus a correction to my earlier comment on `pos_closing_print.py` (posted in that thread: the guard is **not** revertible, it fixes a test that is red on `develop`).

One recalibration worth stating plainly: because the old template called `frappe.db.get_value('POS Profile', ..., 'address')` on a field POS Profile does not have, the previous EOD print raised rather than rendering. **Nothing here is a regression from working behaviour** — this PR restores a broken feature. That lowers the stakes on everything I filed; treat all of it as polish on a net fix, not as blockers.

engahmed1190 (changes requested):
## Requesting changes

To be clear about the bar: this PR is a **net fix** — the old template called `frappe.db.get_value('POS Profile', ..., 'address')` on a field POS Profile does not have, so the previous EOD print raised instead of rendering. Nothing here is a regression. The changes below are asked for because this document is a shift's audit record and two of its empty states are silently wrong, not because the PR makes anything worse.

### Blocking

1. **Payments empty state prints a confidently wrong number.** The `Payments` section title is unconditional while Items and Taxes got `{% if %}` guards, so an empty `payment_reconciliation` prints a dangling header and a Final Summary reading `0.00 / 0.00 / 0.00` with no tag — indistinguishable from a fully reconciled shift. Guard it, or give it an explicit empty-state line. (Thread on the print format.)

2. **Items Sold disappears entirely when empty.** The rewrite dropped the old `{% else %}<div class="muted">No item rows found</div>` fallback, so a lookup that yields nothing prints identically to a genuinely item-less shift. Restore the fallback.

3. **Two identical `if not parent_targets: return []` guards** — the new one at `pos_closing_print.py:90` and the pre-existing one at line 58-59. Keep the new one (the tests pin it; the existing `test_get_items_sold_returns_empty_when_no_transactions` is red on `develop` without it) and delete the inner one.

### Strongly preferred, not blocking

4. Use the stored `p.difference` and `doc.payment_reconciliation | sum(attribute='difference')` rather than recomputing `closing - expected` in the template. `pos_closing_shift.py:73` already computes it precision-rounded; two sources of truth for the same money figure on a printed receipt is worth avoiding, and it deletes the `namespace()` accumulator block.

5. Replace the hardcoded `"{:,.2f}".format(...)` with `frappe.utils.fmt_money(value, currency=...)`. No currency symbol appears anywhere on the report today, and en-US grouping is wrong for the pt-br and id locales this same PR ships translations for.

### Non-blocking

`{:g}` on quantities (goes scientific above 1e6), the grey fills on 80mm thermal stock, the English-only report body under `default_print_language: 'en'`, and the empty PR description. Details in the threads above.

Happy to re-review as soon as 1-3 are in.

pos_next/pos_next/utils/pos_closing_print.py
MohamedAliSmk: removed guard from _fetch_items_for_targets, kept it in get_items_sold

pos_next/pos_next/utils/tests/test_pos_closing_print.py
MohamedAliSmk: Dropped mock_fetch.assert_not_called() from both empty-result tests, and removed the unused _fetch_items_for_targets patches. They now only assert result == [].

pos_next/pos_next/print_format/pos_next_eod_report/pos_next_eod_report.json
MohamedAliSmk: Updated the EOD print format for all five review points:

Difference — uses p.difference and sum(attribute='difference'); removed the namespace() accumulators.
Currency — money via frappe.utils.fmt_money(..., currency=company_doc.default_currency).
Qty / rate — {:.2f} instead of {:g}.
Address — frappe.db.get_value(..., 'address_display') instead of .get_display().
Thermal CSS — dropped grey fills and print-color-adjust: exact; section titles / grands use dashed rules; removed body padding-bottom: 40px.

MohamedAliSmk: All three section headers stay visible now, with explicit empty lines:

Items Sold → No item rows found
Taxes → No tax rows found
Payments → No payment rows found
So a blank lookup no longer looks like a clean, reconciled close.

MohamedAliSmk:
@engahmed1190 
Addressed the three blocking items:

Payments now has an explicit empty state (No payment rows found) instead of a dangling header + misleading 0/0/0 summary.
Items Sold keeps the section header and restores the No item rows found fallback.
Kept the early return in get_items_sold and removed the duplicate guard in _fetch_items_for_targets.
Also applied the preferred template fixes (p.difference / fmt_money) and the thermal/{:g} cleanups. Ready for another look when you have a moment.

MohamedAliSmk:
@engahmed1190 
Addressed the three blocking items:

Payments now has an explicit empty state (No payment rows found) instead of a dangling header + misleading 0/0/0 summary.
Items Sold keeps the section header and restores the No item rows found fallback.
Kept the early return in get_items_sold and removed the duplicate guard in _fetch_items_for_targets.
Also applied the preferred template fixes (p.difference / fmt_money) and the thermal/{:g} cleanups. Ready for another look when you have a moment.

engahmed1190:
## Open items after `9a21409`

The three blocking points are closed — verified by running the tests and rendering the template for both the populated and all-empty cases. Leaving these as a checklist so nothing is lost; **none of them block this PR**, and I'm happy for them to land as follow-ups.

### Worth doing before merge (cheap)

- [x] **Fix the commit message on `9a21409`.** It reads "Kept the empty-target guard in `_fetch_items_for_targets` and removed the duplicate from `get_items_sold`" — the diff does the reverse. The code is correct; the message will mislead anyone reading `git log` later.
- [x] **Fill in the PR description.** Still empty, on a PR whose main diff GitHub cannot render (6.5 KB of Jinja inside a single JSON string). One paragraph on what the report now shows and what was broken before would save the next reader the archaeology.

### Follow-ups

- [x] **Translate the report body.** Every label is hardcoded English ("End of Day Report", "Session Information", "Cashier", "SHORT"/"OVER", the three new "No … rows found" lines) and `default_print_language` is pinned to `"en"`. This PR ships `ar.csv`, `id.csv` and `pt-br.csv` rows for the *button*, so an Arabic cashier gets an Arabic button and an English LTR receipt. Both halves need fixing — `{{ _("...") }}` alone won't translate while `default_print_language: "en"` stands. Arabic also wants `direction: rtl` driven off `frappe.local.lang`.
- [x] **Guard `_fetch_items_for_targets` or document that it must not be called with an empty set.** With the guard now in `get_items_sold`, calling it with an empty set raises `AttributeError: 'NoneType' object has no attribute 'fields_'` from `.where(None)` (confirmed — it's a crash, not an unbounded scan). Its one caller guards, so this is a trap for a future second caller, not a live bug.
- [x] **Escape free-text fields in the template.** Frappe's print-format Jinja env runs with `autoescape=False`, so `{{ it.item_name }}`, `{{ p.mode_of_payment }}`, `{{ t.account_head }}` and `{{ company_doc.company_name }}` render as raw HTML — an item named `<style>*{display:none}</style>` blanks the receipt. Pre-existing, not introduced here. `| e` on those (not on `address_display`, which is intentionally HTML).
- [x] **Extract the template to a sibling `.html` file.** The `html` key being one 6.5 KB JSON line is why both rounds of this review had to be done by extracting and diffing by hand. A `.html` file next to the JSON is diffable, lintable and reviewable in the GitHub UI.
- [x] **Add one render test for the print format.** The Python helper has three tests; the template — every money figure on the shift's audit record — has none. Rendering it against a fixture closing shift and asserting the grand total and the Short/Over string appear would have caught the empty-state issues automatically.

MohamedAliSmk:
> ## Open items after `9a21409`
> 
> The three blocking points are closed — verified by running the tests and rendering the template for both the populated and all-empty cases. Leaving these as a checklist so nothing is lost; **none of them block this PR**, and I'm happy for them to land as follow-ups.
> ### Worth doing before merge (cheap)
> 
>     * [ ]  **Fix the commit message on `9a21409`.** It reads "Kept the empty-target guard in `_fetch_items_for_targets` and removed the duplicate from `get_items_sold`" — the diff does the reverse. The code is correct; the message will mislead anyone reading `git log` later.
> 
>     * [ ]  **Fill in the PR description.** Still empty, on a PR whose main diff GitHub cannot render (6.5 KB of Jinja inside a single JSON string). One paragraph on what the report now shows and what was broken before would save the next reader the archaeology.
> 
> 
> ### Follow-ups
> 
>     * [ ]  **Translate the report body.** Every label is hardcoded English ("End of Day Report", "Session Information", "Cashier", "SHORT"/"OVER", the three new "No … rows found" lines) and `default_print_language` is pinned to `"en"`. This PR ships `ar.csv`, `id.csv` and `pt-br.csv` rows for the _button_, so an Arabic cashier gets an Arabic button and an English LTR receipt. Both halves need fixing — `{{ _("...") }}` alone won't translate while `default_print_language: "en"` stands. Arabic also wants `direction: rtl` driven off `frappe.local.lang`.
> 
>     * [ ]  **Guard `_fetch_items_for_targets` or document that it must not be called with an empty set.** With the guard now in `get_items_sold`, calling it with an empty set raises `AttributeError: 'NoneType' object has no attribute 'fields_'` from `.where(None)` (confirmed — it's a crash, not an unbounded scan). Its one caller guards, so this is a trap for a future second caller, not a live bug.
> 
>     * [ ]  **Escape free-text fields in the template.** Frappe's print-format Jinja env runs with `autoescape=False`, so `{{ it.item_name }}`, `{{ p.mode_of_payment }}`, `{{ t.account_head }}` and `{{ company_doc.company_name }}` render as raw HTML — an item named `<style>*{display:none}</style>` blanks the receipt. Pre-existing, not introduced here. `| e` on those (not on `address_display`, which is intentionally HTML).
> 
>     * [ ]  **Extract the template to a sibling `.html` file.** The `html` key being one 6.5 KB JSON line is why both rounds of this review had to be done by extracting and diffing by hand. A `.html` file next to the JSON is diffable, lintable and reviewable in the GitHub UI.
> 
>     * [ ]  **Add one render test for the print format.** The Python helper has three tests; the template — every money figure on the shift's audit record — has none. Rendering it against a fixture closing shift and asserting the grand total and the Short/Over string appear would have caught the empty-state issues automatically.

@engahmed1190 
Done

PR #300, opened 2026-06-09
fix(shift-closing): count partial-invoice payments in reconciliation

NotAbdelrahmanelsayed:
## Problem

When a cashier collects a payment against an **existing partially-paid / due Sales Invoice** (e.g. 500 cash on an invoice with an outstanding balance), the amount is recorded correctly as a Payment Entry — but it **does not show up in the Close Shift reconciliation**. The cashier ends up with a cash drawer that looks short by exactly the amount collected on due invoices.

## Root cause

A `reference_no` mismatch between how the payment is created and how shift closing queries it:

| | `reference_no` written |
|---|---|
| Normal POS invoice payment | `<opening shift name>` (e.g. `POSA-OS-24-0000001`) |
| Partial / due invoice payment | `POS-<invoice name>` (fallback) |

`get_payments_entries()` in `pos_closing_shift.py` filters `WHERE reference_no = <opening shift name>`, so partial-invoice payments are never matched.

## Fix

Thread the active opening shift from the POS UI down to the Payment Entry, so it is tagged with the shift name and discovered during reconciliation.

**Backend — `pos_next/api/partial_payments.py`**
- `create_payment_entry()` and `add_payment_to_partial_invoice()` accept an optional `pos_opening_shift`.
- When no explicit `reference_no` is supplied, the shift name is used as `reference_no`. The `POS-<invoice>` fallback is preserved for non-shift contexts.

**Frontend**
- `InvoiceManagement.vue` and `PartialPayments.vue` pass the current shift (`useShift().currentShift?.name`) when submitting a payment.

## Result

Cash (and any mode of payment) collected on due/partial invoices during a shift now appears in the Close Shift reconciliation totals.

## Scope

3 files, +16/−1. Backward compatible — `pos_opening_shift` is optional.


NotAbdelrahmanelsayed:
Recreated as #361 since GitHub won't let a bot-closed PR be reopened via the API. Same branch, no code changes — following up there.

PR #301, opened 2026-06-09
feat(cart): add LIFO cart order setting (newest item on top)

NotAbdelrahmanelsayed:
## What

Adds a new POS setting — **LIFO Cart Order (Newest on Top)** — that, when enabled, shows the most recently added item at the **top** of the cart instead of the bottom.

## Why

On long carts, cashiers often want to immediately see the item they just scanned/added for a quick visual confirmation, without scrolling to the bottom. This is a per-profile preference, off by default, so existing behaviour is unchanged.

## How

- **Doctype / backend**: new `cart_lifo` Check field on **POS Settings** (default `0`), registered in `POS_SETTINGS_FIELDS` and `DEFAULT_POS_SETTINGS`.
- **Store**: `posSettings` exposes a `cartLifo` computed.
- **Sorting**: `useCartSort(itemsGetter, lifoMode)` gains an optional `lifoMode`. When **no explicit sort column** is active and LIFO is on, it returns `[...items].reverse()`. Explicit sorts (name/price/quantity/total/addition-order) are completely unaffected.
- **UI**: a toggle in **Settings → Sales Operations**. It lives in that group so saving fires `settings:sales-operations-changed`, which triggers `reloadSettings()` and the cart re-orders live (no page reload).

## Scope

7 files, +36/−4. Default off, fully backward compatible. Production build passes (`vite build` ✓).

> Note: I couldn't attach an in-app screenshot — the test account on the staging site has no POS Profile, so a shift can't be opened to reach the cart/settings UI. The change is verified via a clean production build and isolated diff.


NotAbdelrahmanelsayed:
Recreated as #362 since GitHub won't let a bot-closed PR be reopened via the API. Same branch, no code changes — following up there.

PR #309, opened 2026-06-12
feat(pos): configurable idle auto-focus for item search

NotAbdelrahmanelsayed:
## Summary

Makes the idle "return focus to the item search" behavior a **native, configurable POS setting**, replacing reliance on the external `easy_entry` app's hardcoded `pos_guard.js` (fixed 3s, not configurable).

### New POS Settings (Advanced)
- **`enable_idle_refocus`** (Check, default **1**) — preserves today's behavior out of the box.
- **`idle_refocus_seconds`** (Int, default **3**, `depends_on` enable_idle_refocus) — idle seconds before refocus.

Wired through every layer so the value actually reaches the SPA: doctype JSON, `constants.py` (`POS_SETTINGS_FIELDS` **and** `DEFAULT_POS_SETTINGS`), the `posSettings` store (defaults + computeds + exports), and the settings dialog (checkbox + conditional seconds field).

### Implementation
New composable `useIdleRefocus.js`:
- After the configured idle interval, refocuses the item search input.
- Guards ported from `pos_guard.js` `userIsBusy`: never steals focus from another text field (customer search, quantity edit) and never fights an open dialog (`uiStore.isAnyDialogOpen`).
- Reacts live to the settings (enable toggle + interval) and cleans up on unmount.
- Wired in `POSSale.vue`, which owns `itemsSelectorRef` and the UI store.

### Tests & i18n
- `useIdleRefocus.test.js` (5 cases): fires after the interval, stays off when disabled, holds off while a dialog is open, doesn't steal focus from another input, and re-evaluates the interval reactively.
- Arabic (`ar.csv`) entries for the two new setting labels/descriptions.

### Verification
- `npm run test:run` — all pass. `npm run build` — compiles cleanly.
- Manual screenshot review against a live backend: confirmed the settings checkbox + conditional seconds field render and toggle correctly; verified the item search refocuses ~3s after a real activity event, does **not** refocus before the threshold, and does **not** steal focus while typing in the customer search.

### Deployment note
The new doctype fields require `bench migrate` and (for the `constants.py` change) a gunicorn reload on deploy — not run here against the shared site.

### Follow-up (separate repo)
`easy_entry`'s `pos_guard.js` still targets the SPA and now duplicates this behavior; retiring/disabling it there is a follow-up.


NotAbdelrahmanelsayed:
Recreated as #364 since GitHub won't let a bot-closed PR be reopened via the API. Same branch, no code changes — following up there.

PR #311, opened 2026-06-12
fix(pos): honor scanned barcode UOM in item selection dialog

NotAbdelrahmanelsayed:
## Problem

When a barcode is tied to a UOM that differs from the item's stock UOM — e.g. a barcode keyed to **Nos** on an item stocked in **grams** — scanning it adds the item in the **stock UOM**, not the barcode's UOM.

`search_by_barcode` already resolves this correctly: it sets `item.uom` from `Item Barcode.uom` and returns the right `conversion_factor`. But when the scan opens the UOM dialog (scanner mode without auto-add), `ItemSelectionDialog.loadOptions()` defaulted the selected option to `options[0]` (the stock UOM) whenever there was more than one UOM. Its single-barcode auto-select couldn't help either, because `search_by_barcode` doesn't populate `barcode_uoms` on its response — so the dialog always landed on the stock UOM.

## Fix

When defaulting the UOM selection, prefer an explicitly resolved UOM:

`resolved_uom` (weighted/priced barcodes) → **scanned barcode UOM** (`item.uom` when it differs from `stock_uom`) → a sole `barcode_uoms` entry → stock UOM.

No behaviour change when `item.uom` is absent or equals the stock UOM (normal grid taps, single-UOM items), so existing flows are unaffected.

## Test plan

- Item stocked in gram, barcode keyed to Nos (1 Nos = N g). Scan in scanner mode → UOM dialog now defaults to **Nos** (was gram).
- Normal item tap (no barcode UOM) → still defaults to stock UOM.
- Weighted/priced barcode (`resolved_uom`) → unchanged.


NotAbdelrahmanelsayed:
Recreated as #365 since GitHub won't let a bot-closed PR be reopened via the API. Same branch, no code changes — following up there.

PR #320, opened 2026-06-15
fix: enforce one-time-per-customer offers (server gate + offline/online sync)

engahmed1190:
## Summary

PR #319 introduced the **One-Time Offer per Customer** feature, but a review found the limit **was not actually being enforced** — online or offline. This PR fixes the server-side gate, closes a transaction-rule bypass, makes the offline/online frontend consistent, and hardens the recording hook so it can never break a sale.

No schema change and no data migration — the original `One Time Customer Offer Usage` doctype and its `format:{customer}:{pricing_rule}` autoname are kept. All lookups now use field filters instead of reconstructing the composite name.

---

### 🔴 1. Server gate never matched recorded redemptions
`apply_offers` checked existence with a **double-colon** key — `f"{customer}::{record.name}"` — but the doctype autoname generates a **single-colon** name (`CUST:RULE`). The lookup never matched, so a customer could redeem the same one-time offer **unlimited times online**. Proven on a live site.

**Fix:** field-filter lookup `frappe.db.exists("One Time Customer Offer Usage", {"customer": customer, "pricing_rule": record.name})` — immune to the autoname format.

### 🔴 2. Transaction-scoped one-time offers bypassed the gate entirely
The transaction-rule top-up loop didn't select `one_time_per_customer` and applied no gate, so cart-level one-time offers applied on **every** invoice, including walk-ins and repeat customers.

**Fix:** added the field to the query and routed both loops through a shared `_skip_one_time_rule`.

### 🔴 3. Offline redemption recording was silently dropped
A refactor (commit `cf035dc2`) removed the call that caches a redemption on offline checkout, leaving the whole `recordOfflineRedemptions` chain orphaned. The same customer could re-apply a one-time offer on every offline sale; on sync the server strips the discount, causing a **receipt/cash vs. server grand_total mismatch**.

**Fix:** recording restored in the real offline checkout path (`POSSale.vue`), keyed by the offline invoice id.

### 🟠 4. Walk-in / return sales recorded redemptions
`update_invoice` stamped `pos_applied_one_time_rules` unconditionally, which would write a usage row against the **shared default/walk-in customer** (blocking everyone) or consume a redemption on a credit note.

**Fix:** stamp only for an identified, non-default, non-return customer via the new `is_one_time_eligible_customer` helper.

---

### Backend — `pos_next/api/`
- **`invoices.py`**
  - `apply_offers`: field-filter existence check (#1); shared `_skip_one_time_rule` applied in both the per-item and transaction loops (#2); customer's redeemed set fetched **once** into a `set` (1 query instead of one `exists()` per rule).
  - `update_invoice`: walk-in/return guard on stamping (#4).
  - New `is_one_time_eligible_customer(customer, default_customer, is_return)` — single source of truth shared by the grant side and record side.
- **`sales_invoice_hooks.py`**
  - `record_one_time_offer_usage`: never fails the sale — guards an over-long composite name and logs-and-skips any insert error; idempotency via `ignore_if_duplicate`. Removed the redundant per-rule pre-skip.

### Frontend — `POS/src/`
- **`utils/offline/db.js`** — redemption cache row reshaped to `{ serverRules, offlineRules: { [invoiceId]: [...] } }`:
  - `serverRules` — authoritative, **replaced wholesale** on each online fetch so a server-side release self-heals.
  - `offlineRules` — per-invoice unsynced redemptions, so a voided offline sale releases **only its own**.
  - Backward-compatible with legacy flat `rules` rows. Shared `_putRedemptionRow` read-mutate-write helper; `releaseOfflineRedemptions` returns the effective set.
- **`stores/posOffers.js`** — online fetch is authoritative (replace, preserve offline); `recordOfflineRedemptions` keyed by invoice id; new `releaseOfflineRedemptionsForInvoice` refreshes the in-memory gate without a re-read.
- **`stores/posCart.js`** — `recordOfflineRedemptionsForSale` computes applied one-time rules and delegates to the offers store.
- **`stores/posSync.js`** — `deletePending` releases redemptions on void; new **single-chokepoint** `supersedeInvoice` bundles worker-supersede + release.
- **`utils/offline/sync.js`** — `markInvoiceSynced` releases the offline-bucket entry once synced, **keyed by customer** (no full-table scan per invoice on the sync hot path).
- **`pages/POSSale.vue`** — records redemptions on offline checkout; routes both supersede sites through `posSync.supersedeInvoice`.

### Offline ⇄ online parity matrix
| Event | Behavior |
|---|---|
| Offline sale | Redemption cached locally (keyed by offline id) → blocks next offline sale |
| Online sale | Recorded server-side on submit; gate self-heals on next customer-select |
| Offline void / delete | Cached redemption released → customer not wrongly blocked |
| Offline edit (supersede) | Original's redemption released; replacement records its own |
| Sync to server | Offline-bucket entry dropped; repopulated as authoritative `serverRule` |
| Server-side cancel/release | Self-heals on next online fetch (server set replaced) |

---

## Tests
- New `POS/src/utils/offline/__tests__/oneTimeRedemptions.test.js` — **7 passing** vitest cases covering record (P0), server∪offline union, server-authoritative self-heal (P3), per-invoice release (P2), no-customer scan, release-returns-effective-set, and legacy-row backward compat.
- Backend verified live on `pos-dev`: gate now fires (incl. customer names containing `:`), duplicate insert rejected, field-based lookup correct, long-name checkout degrades gracefully.

### DIFF
--- a/POS/src/pages/POSSale.vue
+++ b/POS/src/pages/POSSale.vue
@@ -2088,9 +2088,13 @@ async function handlePaymentCompleted(paymentData) {
 			// the original row as superseded (keeps audit trail, excludes from sync).
 			if (editingOfflineContext?.originalQueueId) {
 				try {
-					await offlineWorker.supersedeOfflineInvoice(
+					// Supersede also releases the original's cached one-time
+					// redemptions (it never syncs); the new invoice records its own.
+					await offlineStore.supersedeInvoice(
 						editingOfflineContext.originalQueueId,
-						offlineReceiptName
+						offlineReceiptName,
+						editingOfflineContext.originalOfflineId,
+						customerValue
 					);
 				} catch (err) {
 					log.error("Failed to supersede original offline invoice:", err);
@@ -2129,6 +2133,12 @@ async function handlePaymentCompleted(paymentData) {
 			};
 			uiStore.setLastOfflinePrintDoc(offlinePrintDoc);
 			cacheOfflineReceiptPayload(offlineReceiptName, offlinePrintDoc);
+
+			// Cache one-time-per-customer redemptions locally (keyed by this
+			// offline invoice's id) so the next offline sale to this customer
+			// can't reuse the offer. Must run before clearCart() empties the cart.
+			await cartStore.recordOfflineRedemptionsForSale(offlineReceiptName);
+
 			uiStore.showPaymentDialog = false;
 			cartStore.clearCart();
 			// Reset cart hash after successful payment
@@ -2181,9 +2191,14 @@ async function handlePaymentCompleted(paymentData) {
 				if (editingOfflineContext?.originalQueueId) {
 					const serverName = result.name || result.message?.name || null;
 					try {
-						await offlineWorker.supersedeOfflineInvoice(
+						// Supersede also releases the original's cached one-time
+						// redemptions (it never syncs); the online submit just now
+						// recorded the redemption server-side.
+						await offlineStore.supersedeInvoice(
 							editingOfflineContext.originalQueueId,
-							serverName
+							serverName,
+							editingOfflineContext.originalOfflineId,
+							cartStore.customer?.name || cartStore.customer || null
 						);
 					} catch (err) {
 						log.error(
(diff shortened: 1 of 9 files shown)


engahmed1190:
@MohamedAliSmk @MostafaKadry

PR #341, opened 2026-07-28
fix: respect Item Price validity dates when fetching POS prices

tushar-git26:
Item Price queries fetched all rows for an item/price list without filtering on valid_from/valid_upto and collapsed them into a UOM-keyed map where the last arbitrarily-ordered row wins. With multiple Item Price records for the same item, the POS could show an expired or stale price instead of the latest valid one.

Add validity-date filters and order by valid_from so the newest valid price wins, matching ERPNext core get_item_price behaviour. Applies to search_by_barcode, get_item_variants, get_items (grid and template min variant price) and get_items_bulk.

### DIFF
--- a/pos_next/api/items.py
+++ b/pos_next/api/items.py
@@ -390,11 +390,15 @@ def search_by_barcode(barcode, pos_profile):
 		uom_prices = {}
 		if pos_profile_doc.selling_price_list:
 			ItemPrice = DocType("Item Price")
+			today = nowdate()
 			prices = (
 				frappe.qb.from_(ItemPrice)
 				.select(ItemPrice.uom, ItemPrice.price_list_rate)
 				.where(ItemPrice.item_code == item_code)
 				.where(ItemPrice.price_list == pos_profile_doc.selling_price_list)
+				.where(ItemPrice.valid_from.isnull() | (ItemPrice.valid_from <= today))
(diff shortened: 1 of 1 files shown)


MostafaKadry:
Hey! Good catch on the missing `valid_from`/`valid_upto` filtering, but this is already live on `staging`: @MohamedAliSmk  fixed the exact same gap, same 5 call sites in `items.py`. So `staging` has had date-validity filtering on item prices. #341 would just be reintroducing the same fix a second, divergent way.

Diffed both to be sure before saying anything: `86eaa27` pulls the date-range check into one shared `_item_price_validity_conditions()` + `_fetch_uom_prices_map()`/`_fetch_item_uom_prices()` used at all 5 sites, and copies ERPNext core's own predicate verbatim (`IfNull(valid_from,'2000-01-01') <= date & IfNull(valid_upto,'2500-12-31') >= date` . same as `erpnext.stock.get_item_details.get_item_price`). #341 gets to the same result but pastes the null-safe condition inline 5 times instead.

What we will do:

1. **Close #341** as superseded by `86eaa27` (already on staging). no need to carry two versions of the same fix.
2. **Merge `staging` into `develop` soon**, so that `develop` picks up this fix along with everything else currently on staging.

PR #359, opened 2026-09-03
feat: enhance offline batch and serial number management

MohamedAliSmk:
- Added functions to persist and consume batch and serial data for items in offline mode.
- Implemented caching for batch and serial numbers during item searches.
- Improved handling of cached data to ensure accurate updates and retrievals when offline.
- Enhanced error handling for background caching operations.

---

## 🛑 Review action list — do not merge/close until all checked

[Full review](https://github.com/BrainWise-DEV/POSNext/pull/359#pullrequestreview-5142128517) · 11 inline comments · details in each

**Blocking**
- [ ] **T1** Batch qty subtracted twice offline — delete `consumeCachedBatchQty`
- [ ] **T2** Batch consume has no return path (4 drift paths) — fixed by T1
- [ ] **T3** Sold serials returned to cache on offline submit — flag in `clearCart`

**Should fix**
- [ ] **T4** Cache mutators non-atomic — wrap in `db.transaction("rw", …)`
- [ ] **T5** Offline serial fallback skips `isCacheValid` (warehouse + TTL)
- [ ] **T6** Up to 25 sequential POSTs per search — drop pagination sites, dedupe

**Nice to have**
- [ ] **T7** Phantom `db.items` rows + `undefined` warehouse guard
- [ ] **T8** Vitest config + one `items.test.js` (repo has zero tests)
- [ ] **T9** Wrap 118-char import

### DIFF
--- a/POS/src/components/sale/BatchSerialDialog.vue
+++ b/POS/src/components/sale/BatchSerialDialog.vue
@@ -301,7 +301,7 @@ import { Button, Dialog, createResource } from "frappe-ui";
 import { computed, ref, watch } from "vue";
 import { useSerialNumberStore } from "@/stores/serialNumber";
 import { usePOSCartStore } from "@/stores/posCart";
-import { getCachedBatchData, getCachedSerialData } from "@/utils/offline/items";
+import { getCachedBatchData, getCachedSerialData, persistItemBatchSerialData, consumeCachedBatchQty } from "@/utils/offline/items";
 import { isOffline } from "@/utils/offline";
 
 const props = defineProps({
@@ -370,6 +370,13 @@ const batchesResource = createResource({
 				expiry_date: batch.expiry_date,
 				manufacturing_date: batch.manufacturing_date,
 			}));
+
+			// Persist for offline batch selection
+			if (props.item?.item_code) {
+				persistItemBatchSerialData(props.item.item_code, {
+					batch_no_data: data.batch_no_data,
+				}).catch(() => {});
+			}
 		}
 	},
 	onError(error) {
@@ -436,6 +443,8 @@ async function loadBatchesOrSerials() {
 				}));
 				return;
 			}
+			warehouseBatches.value = [];
+			return;
 		}
 		// Fetch from server when online
 		batchesResource.reload();
@@ -447,6 +456,14 @@ async function loadBatchesOrSerials() {
 				availableSerials.value = cachedSerials;
 				return;
 			}
+			// Fallback to in-memory cache from a prior online session
+			const memorySerials = serialStore.getSerials(props.item.item_code);
+			if (memorySerials.length > 0) {
+				availableSerials.value = memorySerials;
+				return;
+			}
+			availableSerials.value = [];
+			return;
 		}
 		// Set warehouse in store
 		serialStore.setWarehouse(props.warehouse);
@@ -496,6 +513,11 @@ function handleConfirm() {
 
 	if (props.item?.has_batch_no && selectedBatch.value) {
 		result.batch_no = selectedBatch.value.batch_no;
+		consumeCachedBatchQty(
+			props.item.item_code,
+			selectedBatch.value.batch_no,
+			props.quantity || 1
+		).catch(() => {});
 	}
 
 	if (props.item?.has_serial_no) {
--- a/POS/src/components/sale/BatchSerialDialog.vue
+++ b/POS/src/components/sale/BatchSerialDialog.vue
@@ -301,7 +301,11 @@ import { Button, Dialog, createResource } from "frappe-ui";
 import { computed, ref, watch } from "vue";
 import { useSerialNumberStore } from "@/stores/serialNumber";
 import { usePOSCartStore } from "@/stores/posCart";
-import { getCachedBatchData, getCachedSerialData, persistItemBatchSerialData } from "@/utils/offline/items";
+import {
+	getCachedBatchData,
+	getCachedSerialData,
+	persistItemBatchSerialData,
+} from "@/utils/offline/items";
 import { isOffline } from "@/utils/offline";
 
 const props = defineProps({
--- a/POS/src/components/sale/BatchSerialDialog.vue
+++ b/POS/src/components/sale/BatchSerialDialog.vue
@@ -456,12 +456,6 @@ async function loadBatchesOrSerials() {
 				availableSerials.value = cachedSerials;
 				return;
 			}
-			// Fallback to in-memory cache from a prior online session
-			const memorySerials = serialStore.getSerials(props.item.item_code);
-			if (memorySerials.length > 0) {
-				availableSerials.value = memorySerials;
-				return;
-			}
 			availableSerials.value = [];
 			return;
 		}
--- a/POS/src/components/sale/BatchSerialDialog.vue
+++ b/POS/src/components/sale/BatchSerialDialog.vue
@@ -301,7 +301,7 @@ import { Button, Dialog, createResource } from "frappe-ui";
 import { computed, ref, watch } from "vue";
 import { useSerialNumberStore } from "@/stores/serialNumber";
 import { usePOSCartStore } from "@/stores/posCart";
-import { getCachedBatchData, getCachedSerialData, persistItemBatchSerialData, consumeCachedBatchQty } from "@/utils/offline/items";
+import { getCachedBatchData, getCachedSerialData, persistItemBatchSerialData } from "@/utils/offline/items";
 import { isOffline } from "@/utils/offline";
 
 const props = defineProps({
@@ -513,11 +513,6 @@ function handleConfirm() {
 
 	if (props.item?.has_batch_no && selectedBatch.value) {
 		result.batch_no = selectedBatch.value.batch_no;
-		consumeCachedBatchQty(
-			props.item.item_code,
-			selectedBatch.value.batch_no,
-			props.quantity || 1
-		).catch(() => {});
 	}
 
 	if (props.item?.has_serial_no) {
--- a/POS/src/stores/itemSearch.js
+++ b/POS/src/stores/itemSearch.js
@@ -1220,6 +1220,12 @@ export const useItemSearchStore = defineStore("itemSearch", () => {
 				// Cache for offline
 				await offlineWorker.cacheItems(items);
 
+				if (shiftStore.profileWarehouse) {
+					cacheBatchSerialForItems(items, shiftStore.profileWarehouse).catch((err) => {
+						log.warn("Background batch/serial caching failed:", err.message);
+					});
+				}
+
 				log.debug(`Fetched page ${page}: ${items.length} items (offset ${start})`);
 			}
 		} catch (error) {
@@ -1328,6 +1334,12 @@ export const useItemSearchStore = defineStore("itemSearch", () => {
 
 				// Cache new batch for offline support
 				await offlineWorker.cacheItems(list);
+
+				if (shiftStore.profileWarehouse) {
+					cacheBatchSerialForItems(list, shiftStore.profileWarehouse).catch((err) => {
+						log.warn("Background batch/serial caching failed:", err.message);
+					});
+				}
 
 				log.debug(`Loaded ${list.length} more items, total: ${totalItemsLoaded.value}`);
 			} else {
@@ -1708,6 +1720,15 @@ export const useItemSearchStore = defineStore("itemSearch", () => {
 
 						// Cache server results for future searches
 						await offlineWorker.cacheItems(serverResults);
+
+						if (shiftStore.profileWarehouse) {
+							cacheBatchSerialForItems(
+								serverResults,
+								shiftStore.profileWarehouse
+							).catch((err) => {
+								log.warn("Background batch/serial caching failed:", err.message);
+							});
+						}
 
 						// If we didn't resolve with cache, resolve with server results
 						if (!cached || cached.length === 0) {
--- a/POS/src/stores/itemSearch.js
+++ b/POS/src/stores/itemSearch.js
@@ -14,6 +14,10 @@ import { useRealtimePosProfile } from "@/composables/useRealtimePosProfile";
 
 const log = logger.create("ItemSearch");
 
+/** Recently fetched batch/serial item codes → timestamp (dedupe search fan-out) */
+const recentlyFetchedBatchSerial = new Map();
+const BATCH_SERIAL_FETCH_TTL_MS = 5 * 60 * 1000;
+
 /**
  * Fetch and cache batch/serial data for items with batch or serial tracking
  * This ensures batch/serial selection works offline
@@ -31,11 +35,23 @@ async function cacheBatchSerialForItems(items, warehouse) {
 		return;
 	}
 
-	log.info(`Caching batch/serial data for ${batchSerialItems.length} items`);
+	const now = Date.now();
+	const itemCodes = batchSerialItems
+		.map((item) => item.item_code)
+		.filter((code) => {
+			const lastFetched = recentlyFetchedBatchSerial.get(code);
+			return !lastFetched || now - lastFetched > BATCH_SERIAL_FETCH_TTL_MS;
+		});
+
+	if (itemCodes.length === 0) {
+		log.debug("All batch/serial items were fetched recently — skipping");
+		return;
+	}
+
+	log.info(`Caching batch/serial data for ${itemCodes.length} items`);
 
 	// Fetch in batches to avoid too large requests
 	const BATCH_SIZE = 20;
-	const itemCodes = batchSerialItems.map((item) => item.item_code);
 
 	for (let i = 0; i < itemCodes.length; i += BATCH_SIZE) {
 		const batchCodes = itemCodes.slice(i, i + BATCH_SIZE);
@@ -51,6 +67,11 @@ async function cacheBatchSerialForItems(items, warehouse) {
 			if (Object.keys(data).length > 0) {
 				await updateItemBatchSerialData(data);
 				log.debug(`Cached batch/serial data for ${Object.keys(data).length} items`);
+			}
+
+			const fetchedAt = Date.now();
+			for (const code of batchCodes) {
+				recentlyFetchedBatchSerial.set(code, fetchedAt);
 			}
 		} catch (error) {
 			log.warn(`Failed to fetch batch/serial data for batch ${i}:`, error.message);
@@ -1220,12 +1241,6 @@ export const useItemSearchStore = defineStore("itemSearch", () => {
 				// Cache for offline
 				await offlineWorker.cacheItems(items);
 
-				if (shiftStore.profileWarehouse) {
-					cacheBatchSerialForItems(items, shiftStore.profileWarehouse).catch((err) => {
-						log.warn("Background batch/serial caching failed:", err.message);
-					});
-				}
-
 				log.debug(`Fetched page ${page}: ${items.length} items (offset ${start})`);
 			}
 		} catch (error) {
@@ -1334,12 +1349,6 @@ export const useItemSearchStore = defineStore("itemSearch", () => {
 
 				// Cache new batch for offline support
 				await offlineWorker.cacheItems(list);
-
-				if (shiftStore.profileWarehouse) {
-					cacheBatchSerialForItems(list, shiftStore.profileWarehouse).catch((err) => {
-						log.warn("Background batch/serial caching failed:", err.message);
-					});
-				}
 
 				log.debug(`Loaded ${list.length} more items, total: ${totalItemsLoaded.value}`);
 			} else {
--- a/POS/src/stores/serialNumber.js
+++ b/POS/src/stores/serialNumber.js
@@ -11,6 +11,11 @@
 import { defineStore } from "pinia";
 import { ref } from "vue";
 import { call } from "@/utils/apiWrapper";
+import {
+	consumeCachedSerials,
+	persistItemBatchSerialData,
+	returnCachedSerials,
+} from "@/utils/offline/items";
 import { logger } from "@/utils/logger";
 
 const log = logger.create("SerialNumber");
@@ -107,6 +112,10 @@ export const useSerialNumberStore = defineStore("serialNumber", () => {
 			});
 
 			log.success(`Loaded ${serials.length} serials for ${itemCode}`);
+
+			// Persist for offline batch/serial selection
+			persistItemBatchSerialData(itemCode, { serial_no_data: serials }).catch(() => {});
+
 			return serials;
 		} catch (error) {
 			log.error(`Failed to fetch serials for ${itemCode}`, error);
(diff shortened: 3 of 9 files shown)


POS/src/components/sale/BatchSerialDialog.vue
```diff
@@ -370,6 +370,13 @@ const batchesResource = createResource({
 				expiry_date: batch.expiry_date,
 				manufacturing_date: batch.manufacturing_date,
 			}));
+
+			// Persist for offline batch selection
+			if (props.item?.item_code) {
+				persistItemBatchSerialData(props.item.item_code, {
```
engahmed1190, line 380: **Nice — this is the actual fix for PN-83.** ✅

The dialog's own batch fetch was never persisted, so `getCachedBatchData()` came back empty offline. Same for `fetchSerials` in the serial store, plus the two new `return` guards below. Those ~30 lines are the whole bug fix and they're correct.

Everything flagged in my other comments is the *consumption-accounting* layer built on top, which wasn't part of PN-83. See the task list in the review summary — the ask is to split it out.

POS/src/components/sale/BatchSerialDialog.vue
```diff
@@ -496,6 +513,11 @@ function handleConfirm() {
 
 	if (props.item?.has_batch_no && selectedBatch.value) {
 		result.batch_no = selectedBatch.value.batch_no;
+		consumeCachedBatchQty(
```
engahmed1190: ### 🔴 T1 — Batch quantity is subtracted **twice** offline

`consumeCachedBatchQty` decrements `batch_qty` in IndexedDB. The item then goes into the cart, and the existing `availableBatches` computed (line 333) *already* subtracts cart quantities from `warehouseBatches`. Offline, `warehouseBatches` is loaded from that same decremented cache — so the same units are removed twice.

Reproduced:

```js
let cache = [{ batch_no: "B1", batch_qty: 10 }];
consumeCachedBatchQty("B1", 2);          // cache -> 8
cart.push({ batch_no: "B1", quantity: 2 });
availableBatches(loadOffline())[0].qty;  // -> 6   (should be 8)
```

Online this is masked because `batchesResource.reload()` overwrites `warehouseBatches` from the server. Offline — the exact scenario this PR targets — stock is under-reported on every batch sale and compounds per transaction, until the cashier is blocked from selling stock that physically exists.

**Suggested fix:** delete `consumeCachedBatchQty` entirely. `availableBatches` is already the single correct mechanism for "warehouse qty minus what's in the cart".

POS/src/components/sale/BatchSerialDialog.vue
```diff
@@ -447,6 +456,14 @@ async function loadBatchesOrSerials() {
 				availableSerials.value = cachedSerials;
 				return;
 			}
+			// Fallback to in-memory cache from a prior online session
+			const memorySerials = serialStore.getSerials(props.item.item_code);
```
engahmed1190: ### 🟡 T5 — Fallback bypasses the store's own validity checks

`serialStore.getSerials()` is a raw `Map` read. It skips `isCacheValid()` (`serialNumber.js:46`), which checks **warehouse match** and a **5-minute TTL**. So serials cached for a different warehouse, or from an hour-old session, are presented to the cashier as current.

Either call through a validity check, or drop this fallback — `getCachedSerialData()` above already covers the offline case now that `fetchSerials` persists.

POS/src/components/sale/BatchSerialDialog.vue
```diff
@@ -301,7 +301,7 @@ import { Button, Dialog, createResource } from "frappe-ui";
 import { computed, ref, watch } from "vue";
 import { useSerialNumberStore } from "@/stores/serialNumber";
 import { usePOSCartStore } from "@/stores/posCart";
-import { getCachedBatchData, getCachedSerialData } from "@/utils/offline/items";
+import { getCachedBatchData, getCachedSerialData, persistItemBatchSerialData, consumeCachedBatchQty } from "@/utils/offline/items";
```
engahmed1190: 🟢 Nit: 118 chars on one line while every other import in this file wraps. Worth splitting for consistency:

```js
import {
	getCachedBatchData,
	getCachedSerialData,
	persistItemBatchSerialData,
	consumeCachedBatchQty,
} from "@/utils/offline/items";
```

POS/src/utils/offline/items.js
```diff
@@ -173,6 +206,70 @@ export const updateItemBatchSerialData = async (batchSerialDataMap) => {
...
+		);
+
+		await persistItemBatchSerialData(itemCode, { serial_no_data: merged });
+	} catch (error) {
+		console.error("Error returning cached serials:", error);
+	}
+};
+
+// Decrease batch quantity in offline cache after selection/sale
+export const consumeCachedBatchQty = async (itemCode, batchNo, qty) => {
```
engahmed1190: ### 🔴 T2 — Batch consumption has no return path, so the drift is permanent *and* arbitrary

Serials are symmetric (`consumeCachedSerials` ↔ `returnCachedSerials`, wired into remove / qty-edit / clear). Batches have consume only. Four ways the cache diverges with nothing to correct it:

1. `useInvoice.js:339` — removing a batch line returns serials, never batch qty.
2. `useInvoice.js:380` — `updateItemQuantity` adjusts serials on qty change; batches aren't handled at all. Confirm at qty 2, edit to 5 → 2 consumed, 5 in cart.
3. `useInvoice.js:220` — `addItem` dedupes on `item_code + uom`, **not** `batch_no`. Picking a second batch for an item already in the cart merges into the existing line and silently drops the new `batch_no` — but this function already decremented it. Stock vanishes for a batch that never entered the cart.
4. `POSSale.vue:2581` — `cartStore.addItem` is inside a `try/catch`; on throw the qty is consumed and nothing is added.

The only thing that heals this is a re-sync (`cacheItems` `bulkPut` replaces the whole row and drops `batch_no_data`), which is accidental, not designed.

Removing this function per T1 resolves all four at once.

POS/src/utils/offline/items.js
```diff
@@ -173,6 +206,70 @@ export const updateItemBatchSerialData = async (batchSerialDataMap) => {
 	}
 };
 
+// Remove consumed serial numbers from offline cache
+export const consumeCachedSerials = async (itemCode, serialNumbers) => {
```
engahmed1190, line 211: ### 🟡 T4 — All four cache mutators are non-atomic read-modify-write

`consumeCachedSerials`, `returnCachedSerials`, `consumeCachedBatchQty` and `updateItemBatchSerialData` each do `get` → compute → `db.items.update()` with no `db.transaction("rw", db.items, ...)`.

Two concurrent consumes, or a consume racing the background `cacheBatchSerialForItems` write for the same item, silently loses one update. And every call site is fire-and-forget with `.catch(() => {})`, so it's completely invisible when it happens.

```js
await db.transaction("rw", db.items, async () => {
	const serials = await getCachedSerialData(itemCode);
	...
});
```

Applies to whichever of these survives the split.

POS/src/utils/offline/items.js
```diff
@@ -146,20 +146,53 @@ export const getCachedSerialData = async (itemCode) => {
...
+		const update = {};
+		if (data.batch_no_data !== undefined) update.batch_no_data = data.batch_no_data;
+		if (data.serial_no_data !== undefined) update.serial_no_data = data.serial_no_data;
+
+		if (Object.keys(update).length === 0) return false;
+
+		const item = await db.items.get(itemCode);
+		if (!item) {
+			// Item not in cache yet — store batch/serial data so offline selection still works
+			await db.items.put({ item_code: itemCode, ...update });
```
engahmed1190: ### 🟢 T7 — This can create phantom item rows

`put` with only `{ item_code, batch_no_data }` writes a row with no `item_name`, `barcodes`, price or item group. That row is then returned by `getCachedItems` / `searchCachedItems` — the offline grid renders from `offlineWorker.searchCachedItems("", …)` — and counted in cache stats. Result: a blank tile in the product grid.

The trigger is narrow (both current callers work on items already in cache), so this is low severity, but it's a silent data-shape violation. Prefer skipping the write when the item is missing, or seed the row from the caller's item object.

POS/src/utils/offline/items.js
```diff
@@ -173,6 +206,70 @@ export const updateItemBatchSerialData = async (batchSerialDataMap) => {
...
+export const returnCachedSerials = async (itemCode, serialNumbers) => {
+	try {
+		if (!itemCode) return;
+
+		const serials = await getCachedSerialData(itemCode);
+		const toReturn = parseSerialNumbers(serialNumbers);
+		if (!toReturn.length) return;
+
+		const existing = new Set(serials.map((s) => s.serial_no));
+		const warehouse = serials[0]?.warehouse;
```
engahmed1190: 🟢 T7b — If the cache was emptied by prior consumption, `serials` is `[]`, so `warehouse` is `undefined` and every returned serial is written unscoped. Worth guarding or taking the warehouse as a parameter.

POS/src/stores/serialNumber.js
```diff
@@ -168,6 +179,8 @@ export const useSerialNumberStore = defineStore("serialNumber", () => {
 			a.serial_no.localeCompare(b.serial_no, undefined, { numeric: true })
 		);
 
+		returnCachedSerials(itemCode, serialNumbers).catch(() => {});
```
engahmed1190: ### 🔴 T3 — Serials sold offline get written back into durable storage as available

Not introduced here, but **this line is what makes it bite.**

`clearCart` (`useInvoice.js:1182`) returns every serial to the cache, and `POSSale.vue:2195` is the **offline payment success** path — it can't distinguish "cart abandoned" from "cart sold". Previously `returnSerials` only touched a 5-minute in-memory `Map`, so a page reload cleared it. Now it writes IndexedDB, and the offline dialog reads IndexedDB *first* (`loadBatchesOrSerials`, before the in-memory fallback). The just-sold serial is offered again on the next invoice and survives a reload.

Offline invoices sync via `pos_next.api.invoices.submit_invoice`, so ERPNext rejects the duplicate serial — the outcome is **failed syncs and manual reconciliation at end of shift**, not a silent double-sell. Still needs fixing.

**Fix belongs in `clearCart`**, not here: take a `returnSerials` flag (or a `reason`) and pass `false` from the two post-submit call sites.

POS/src/stores/itemSearch.js
```diff
@@ -1709,6 +1721,15 @@ export const useItemSearchStore = defineStore("itemSearch", () => {
 						// Cache server results for future searches
 						await offlineWorker.cacheItems(serverResults);
 
+						if (shiftStore.profileWarehouse) {
+							cacheBatchSerialForItems(
```
engahmed1190, line 1734: ### 🟡 T6 — Up to 25 sequential POSTs per server search

`cacheBatchSerialForItems` chunks by 20 and `await`s **sequentially**, and server-side `get_batch_serial_data_for_items` runs a per-item `get_batch_qty()` loop.

`searchBatchSize` defaults to 500, so a search where most results are tracked items = **25 sequential round-trips**, each doing 20 stock-ledger aggregations. `searchItems` is debounced 300ms so this is per *search*, not per keystroke — but three searches in a row queue three of these and nothing cancels an in-flight run.

It's also largely redundant: `startBackgroundCacheSync` (line 1641) already fans out over the first 10,000 cached items at shift start.

**Suggested:** keep this one site if you want fresh-on-demand coverage, but add a `Set` of recently-fetched item codes with a short TTL so the same items aren't re-fetched on every search.

POS/src/stores/itemSearch.js
```diff
@@ -1220,6 +1220,12 @@ export const useItemSearchStore = defineStore("itemSearch", () => {
 				// Cache for offline
 				await offlineWorker.cacheItems(items);
 
+				if (shiftStore.profileWarehouse) {
+					cacheBatchSerialForItems(items, shiftStore.profileWarehouse).catch((err) => {
```
engahmed1190: ### 🟡 T6b — Pagination sites are redundant with the shift-start sync

This and the `loadMoreItems` site (line 1339) fire the same 20-per-request sequential fan-out on every page of browsing. `startBackgroundCacheSync` (line 1641) has already cached batch/serial data for the first 10k items — these only add value for items that appear *after* that sync completes.

Suggest dropping both pagination sites and keeping only the search-path one, deduped.

---

🟢 Separate, minor: `await offlineWorker.cacheItems(items)` on the line above uses `bulkPut`, which replaces the whole row and **drops** `batch_no_data`/`serial_no_data`. This un-awaited call restores it milliseconds later. Going offline inside that window leaves the item with no batch data — precisely the bug this PR fixes. Merging the batch data into the put would close it.

engahmed1190 (changes requested):
## 🔍 Full review — code logic, security, performance, coverage

**Verdict: request changes.** The offline *read* fix is correct and should ship. The *consumption-accounting* layer bolted on top has three high-severity defects and wasn't part of PN-83.

### Structural read

What actually fixes "Batch/Serial unable to select at offline mode" is ~30 lines:

| Change | Why it fixes PN-83 |
|---|---|
| `persistItemBatchSerialData` in `batchesResource.onSuccess` | the dialog's own batch fetch was never written to IndexedDB |
| `persistItemBatchSerialData` in `fetchSerials` | serial store results were never written to IndexedDB |
| the two new `return` guards in `loadBatchesOrSerials` | offline previously fell through to `batchesResource.reload()`, which fails |

Everything else — the three new `cacheBatchSerialForItems` call sites, `consumeCachedBatchQty`, `consumeCachedSerials`, `returnCachedSerials`, and the create-on-missing `db.items.put` — is offline inventory accounting. Every defect below lives in that second layer. Cutting it removes ~90 of the 161 added lines **and every high-severity finding**.

### Security

**No new attack surface.** No new endpoints, no new server-side code, no HTML/URL/query sink. `get_batch_serial_data_for_items` is `@frappe.whitelist()` and unchanged by this diff. All new code writes client-side IndexedDB, already trusted-per-device here.

The security-relevant findings *are* the inventory-integrity ones: T1/T2 understate stock, T3 lets the same serial reach two invoices. For a POS that's the risk surface that matters.

### Coverage

**Zero.** `find . -name "*.test.*"` returns nothing, and there's no `test` block in `POS/vite.config.js` despite `test` / `test:run` / `test:coverage` in `package.json` — `yarn test` cannot run today. This diff adds four pure-ish mutators with real arithmetic (set diffing, `Math.max(0, qty - n)`, sort-merge); they're the most testable and highest-consequence code in the changeset.

### Verified

- Patch applies cleanly to current `develop` (`git apply --3way`) — no conflicts, no duplicated call sites.
- `npx biome lint` on the changed files: no new violations (one pre-existing `useOptionalChain` at `BatchSerialDialog.vue:365`).
- T1 reproduced with a standalone script — see the inline comment.

---

## ✅ Task list — please do not merge or close this PR until every box is checked

### 🔴 Blocking

- [ ] **T1** — Batch qty subtracted twice offline. Delete `consumeCachedBatchQty`; `availableBatches` already handles cart subtraction. `BatchSerialDialog.vue:516`
- [ ] **T2** — Batch consumption has no return path (4 divergence paths: line remove, qty edit, `addItem` uom-merge dropping `batch_no`, `addItem` throw). Resolved by T1. `items.js:254`
- [ ] **T3** — Sold serials returned to durable cache on offline submit. Fix in `clearCart` (`useInvoice.js:1182`) with a `returnSerials` flag, `false` from the two post-submit call sites. `serialNumber.js:182`

### 🟡 Should fix

- [ ] **T4** — Wrap the surviving cache mutators in `db.transaction("rw", db.items, ...)`; they're non-atomic read-modify-write with `.catch(() => {})` at every call site. `items.js:210`
- [ ] **T5** — Offline serial fallback bypasses `isCacheValid` (warehouse + 5-min TTL). `BatchSerialDialog.vue:460`
- [ ] **T6** — Drop the two pagination `cacheBatchSerialForItems` sites; keep the search one with a recently-fetched `Set` + TTL. Up to 25 sequential POSTs per search today. `itemSearch.js:1224`, `1339`, `1725`

### 🟢 Nice to have

- [ ] **T7** — Don't `put` phantom item rows (no `item_name`/`barcodes`); guard the `undefined` warehouse in `returnCachedSerials`. `items.js:173`, `236`
- [ ] **T8** — Add a `test` block to `POS/vite.config.js` + one `items.test.js` with `db` mocked, covering the consume/return round-trip, `Math.max(0, …)` clamping, and `parseSerialNumbers` on both string and array input. Not a suite — one file.
- [ ] **T9** — Wrap the 118-char import at `BatchSerialDialog.vue:304`.

### Latent / no action needed this PR

- Persisted batch/serial data has no warehouse key. Harmless today (dialog is bound to `shiftStore.profileWarehouse`, `POSSale.vue:613`) but a shift/profile change leaves stale data with nothing to invalidate it. Worth an issue.

---

https://claude.ai/code/session_016WM1HqLHEhF8GhYjB4EJHp

POS/src/components/sale/BatchSerialDialog.vue
MohamedAliSmk: Agreed — kept the PN-83 read path intact (`persistItemBatchSerialData` on batch/serial fetch + offline `return` guards). The rest of this series only trims/fixes the consumption-accounting layer per your task list.

POS/src/components/sale/BatchSerialDialog.vue
MohamedAliSmk: Fixed in `4b5eb310` — removed the `consumeCachedBatchQty` call from `handleConfirm`. `availableBatches` remains the single source of truth for warehouse − cart.

POS/src/utils/offline/items.js
MohamedAliSmk: Fixed in `59e6f51a` — deleted `consumeCachedBatchQty` entirely. That closes all four divergence paths you listed (line remove, qty edit, uom-merge dropping `batch_no`, addItem throw).

POS/src/stores/serialNumber.js
MohamedAliSmk: Fixed in `65fba4f3` — `clearCart({ returnSerials = true })` in `useInvoice` / `posCart`. Both post-submit call sites in `POSSale.vue` (offline + online payment success) pass `returnSerials: false`. Abandoned clears / draft save / manual clear still restore serials.

POS/src/utils/offline/items.js
MohamedAliSmk: Fixed in `b2adb1cf` — wrapped surviving mutators (`persistItemBatchSerialData`, `updateItemBatchSerialData`, `consumeCachedSerials`, `returnCachedSerials`) in `db.transaction("rw", db.items, ...)`.

POS/src/components/sale/BatchSerialDialog.vue
MohamedAliSmk: Fixed in `ddf82e24` — dropped the in-memory `getSerials()` fallback. Offline now relies on `getCachedSerialData()` only (populated by `fetchSerials` persist).

POS/src/stores/itemSearch.js
MohamedAliSmk: Fixed in `bd6f34d4` — kept the search-path site and added a 5-minute `recentlyFetchedBatchSerial` Map so the same item codes aren’t re-fetched on every search.

POS/src/stores/itemSearch.js
MohamedAliSmk: Fixed in `bd6f34d4` — removed both pagination call sites (`fetchItems` page load + `loadMoreItems`). Shift-start sync + deduped search path remain. Noted the `cacheItems` bulkPut race for a follow-up; not closed in this series.

POS/src/utils/offline/items.js
MohamedAliSmk: Fixed in `3d879240` — if the item row is missing we skip the write instead of `put`ting a phantom `{ item_code, batch/serial }` row.

POS/src/utils/offline/items.js
MohamedAliSmk: Fixed in `3d879240` — `returnCachedSerials(itemCode, serials, warehouse)` now takes warehouse from the caller (`serialStore.currentWarehouse`); if neither param nor cache has a warehouse, we no-op.

POS/src/components/sale/BatchSerialDialog.vue
MohamedAliSmk: Fixed in `438f9ddb` — import wrapped multi-line (without `consumeCachedBatchQty`, which left in T1/T2).

MohamedAliSmk:
## Review follow-up (local, not pushed yet)

Addressed T1–T9 as **9 separate commits** on `PN-83-Batch-Serial-unable-to-select-at-offline-mode`. Inline replies are on each thread.

| Commit | Point |
|---|---|
| `4b5eb310` | T1 — remove `consumeCachedBatchQty` call |
| `59e6f51a` | T2 — delete `consumeCachedBatchQty` |
| `65fba4f3` | T3 — `clearCart({ returnSerials: false })` post-submit |
| `b2adb1cf` | T4 — Dexie transactions on surviving mutators |
| `ddf82e24` | T5 — drop invalid in-memory serial fallback |
| `bd6f34d4` | T6 — drop pagination sites; TTL-dedupe search |
| `3d879240` | T7 — no phantom rows; warehouse guard on return |
| `93d58875` | T8 — vite `test` block + `items.test.js` (5 tests green) |
| `438f9ddb` | T9 — wrap offline items import |

**T8 note:** `Math.max(0, …)` clamping lived only in `consumeCachedBatchQty`, which left with T1/T2 — tests cover consume/return round-trip, `parseSerialNumbers`, phantom skip, and warehouse guard instead.

Holding push until author review of the local commits.

POS/src/utils/offline/items.js
MohamedAliSmk: Fair point — fire-and-forget `.catch(() => {})` on non-atomic RMW made lost updates invisible.

**Done in `b2adb1cf`:** wrapped the surviving mutators in `db.transaction("rw", db.items, …)`:
- `persistItemBatchSerialData`
- `updateItemBatchSerialData`
- `consumeCachedSerials`
- `returnCachedSerials`

(`consumeCachedBatchQty` was already removed by T1/T2.) Nested calls reuse Dexie’s same-table transaction.

MohamedAliSmk:
### T8 — tests

Agreed the mutators were the most testable / highest-consequence part of the changeset, and `yarn test` couldn’t run without a Vite `test` block.

**Done in `93d58875`:**
- added a `test` block to `POS/vite.config.js` (`jsdom`, `src/**/*.{test,spec}.{js,ts}`)
- added `POS/src/utils/offline/items.test.js` with `db` mocked

Coverage included:
- `parseSerialNumbers` on string + array (+ empty)
- consume → return round-trip
- phantom-row skip when item is missing
- warehouse guard when cache is empty

Note: `Math.max(0, …)` clamping lived only inside `consumeCachedBatchQty`, which was removed in T1/T2, so that case isn’t tested anymore.

`yarn test:run src/utils/offline/items.test.js` → 5 passed.

PR #318, opened 2026-06-15
Pn 65 expenses in pos screen

MohamedAliSmk:
## Summary
- Added ExpenseDialog component for recording expenses in the POS.
- Updated ShiftClosingDialog to display POS expenses summary and net cash impact.
- Enhanced InvoiceCart to include a button for recording POS expenses.
- Introduced new computed properties for managing expenses in the POS shift store.
- Updated translations to support new expense-related features.

## Security decision: `ignore_permissions` on Employee / Account lookups

POS cashiers typically lack **Employee** and **Account** read permissions, so the till cannot populate the expense dialog with a normal `get_list` under their role.

`get_active_employees` and `get_expense_accounts` therefore use `ignore_permissions=True` **only after** `validate_open_shift` proves the caller owns an open shift for the requested profile. That is an intentional till trade-off:

| Control | Behaviour |
|--------|-----------|
| Gate | Open-shift ownership (`validate_open_shift`) before any bypassed read |
| Scope | Shift company only; Active employees; non-group, non-disabled expense ledgers |
| Bound | Accounts capped (`EXPENSE_ACCOUNT_PAGE_LENGTH`, default 50) with `search_expense_accounts` for further search |
| Fields | Employees return only `name` and `employee_name` (what the dialog select needs) |
| Revisit | Prefer a narrower Employee/Account role for cashiers if HR/chart exposure at the till becomes unacceptable |

Also recorded in the module docstring of `pos_next/api/expenses.py`.

## Test plan
- [ ] Open a shift as a cashier without Employee/Account read perms; expense dialog still loads accounts and employees
- [ ] Account autocomplete searches server-side and does not dump the full chart
- [ ] Recording an expense creates a submitted Journal Entry and updates shift totals
- [ ] Closing shift shows expenses / net cash impact correctly

## N3 — branch narrowed (force-push)

Unrelated riders were dropped from this PR so the diff is expenses-only:

| Dropped | Lands via |
|--------|-----------|
| `8231e26` offline batch/serial | [#359](https://github.com/BrainWise-DEV/POSNext/pull/359) |
| `05a7b28` / `59acd2f` / `9e9c26f` EOD + translations | [#275](https://github.com/BrainWise-DEV/POSNext/pull/275) |
| `f5ae353` collected vs invoiced (+ undo commit) | already on `develop` via [#312](https://github.com/BrainWise-DEV/POSNext/pull/312) |
[...]

PR #360, opened 2026-09-03
Feature/thin develop split

MohamedAliSmk:


engahmed1190 (commented):
## Review: architecture, security, performance, coverage

Reviewed against `5d4c16f` (merge base `a07f981`). Verified on a local bench: 3 sites, `dns_multitenant: true`, 9 gunicorn workers, Frappe + ERPNext v16-era Sales Invoice. JS suite run (`vitest run` — 12 pass), `compileall` clean, ruff and biome run on the changed files.

**One blocker, four high, nine medium.** The blocker is that the split's central mechanism does not work: the optional-app gate is evaluated once per worker process, against whichever site that worker happened to touch first. On this bench a site that does not even have `pos_next` installed decided `pos_next`'s pricing-rule patch state for a site that does. Details inline on `pos_next/__init__.py`.

The authorization gate is well-built underneath — the registry/gate/grants/policy separation is clean, `consume` being compare-and-set rather than delete is the right call and the reasoning in the docstring is correct, and it is the best-tested code in the PR. The problems are at its edges: the approver dialog never shows what is being approved, the rate limit keys on IP rather than approver, PIN enrolment is self-service, and the two audit fields it writes to the invoice are never created.

### Not blocking, but worth deciding before merge

- `promoApi.js` routes authorization to `posnext_promotions.api.authorization.*` whenever that app is installed, while `before_submit` is still enforced by `pos_next`'s gate reading `pos_next`'s redis keyspace. Nothing in this repo can verify that pairing and no test covers it.
- `applyOfflineGiftPool` / `applyOfflineGwpSameItem` (~200 lines of satellite behaviour) are hardcoded in `posCart.js` and gated at runtime — while this same PR ships `offerStrategies.js`, a plugin registry built for exactly that.

### Coverage

`authorization/` is genuinely well tested for its size. The gaps are where the money is: `overrides/sales_invoice.py` (the largest money-path rewrite here) and `posCart.js` (+396) have zero tests, and four of the findings below live in that untested code. `test_no_promotions_import.py` is solid — a real AST walk. `test_split_smoke.py` mostly `skipTest`s when the optional app is absent, and a couple of cases assert nothing that can fail.

**Does POS Next work standalone?** Static imports: yes, verified — nothing imports `posnext_promotions`, `magento_integration` or `masar_miraaya`, and the app imports and patches cleanly without them. Runtime: not established, for the two reasons above.

### One correction to an earlier draft of this review

I initially flagged `get_invoices` as having lost its search/date filters and its page-size clamp. That was my error: I diffed against `develop`'s tip rather than the merge base. `98b39e5` improved `get_invoices` on `develop` *after* this branch point; this PR never touched it and the merge will take develop's version. Withdrawn.

*Filed as comments, not a change request — the call on merge is yours.*

pos_next/__init__.py
```diff
@@ -12,42 +12,49 @@ def console(*data):
...
+
+
 # Patch get_other_conditions to exclude pos_only pricing rules from non-POS documents.
 # No Frappe hook exists for non-whitelisted module-level functions (override_whitelisted_methods
 # only works for @frappe.whitelist() HTTP endpoints, override_doctype_class only for DocType
 # classes). This is the standard Python module init approach — runs once at import.
-try:
-	from erpnext.accounts.doctype.pricing_rule import utils as pr_utils
+# Skip when posnext_promotions is installed so that app's patches are the single owner.
+if not _promotions_installed():
```
engahmed1190: **BLOCKER — this gate is decided once per worker process, from an arbitrary site.**

`_promotions_installed()` calls `frappe.get_installed_apps()` at module import. `pos_next` is imported during `frappe.init()`, which happens once per Python process, so this decision is frozen against whichever site that worker touched first and then applies to every other site it serves.

Reproduced on a 3-site bench:

```
f.init(site='fleet.localhost');   f.destroy()
f.init(site='posnext.local');     f.destroy()
f.init(site='service.localhost')

-> get_installed_apps called during import, once per process:
     ('fleet.localhost', ['frappe','erpnext','brainwise_fleet','nexmove_service','payments'])
   total calls: 1
```

`fleet.localhost` does not have `pos_next` installed at all, yet it decided the patch state for `posnext.local`. With `dns_multitenant: true` and 9 gunicorn workers:

- worker warmed by a site **with** `posnext_promotions` -> patches disabled for every site without it (silent wrong prices);
- worker warmed by a site **without** it -> double-patching where both apps own the same override;
- which worker serves a request is arbitrary, so the same cart can price differently on consecutive requests.

Outside a site context the call raises and is swallowed to `False` (`RuntimeError: object is not bound`), so `bench build` and other non-site imports always patch.

**Fix:** drop the import-time gate and guard inside each patched function at call time — exactly how `optional_apps.promotions_installed()` is already used at request time in `overrides/pricing_rule.py` and `api/sales_invoice_hooks.py`. Then delete `_promotions_installed()` from here: it duplicates that helper, so a fix to one silently misses the other.

pos_next/api/wallet.py
```diff
@@ -142,36 +152,73 @@ def get_wallet_amount_from_payments(payments):
...
+	if payments is None:
+		payments = frappe.get_all(
+			"Sales Invoice Payment",
+			filters={"parent": invoice_name},
+			fields=["mode_of_payment", "amount"],
+		)
+	return get_wallet_amount_from_payments(payments)
+
+
+def _get_wallet_payment_modes():
```
engahmed1190, line 186: **High — this cache has no TTL and no invalidation, and it gates a balance check.**

`frappe.cache().set_value(...)` with no expiry, and no `on_update` hook on Mode of Payment clears it. This replaced a per-payment `db.get_value` that was correct.

Failure: an admin ticks `is_wallet_payment` on a Mode of Payment after the cache is warm. `get_wallet_amount_from_payments` returns 0 -> `validate_wallet_payment` returns early -> the balance check never runs -> a customer can spend more wallet balance than they hold, until someone flushes redis.

Add a TTL, or clear the key from a Mode of Payment `on_update`.

pos_next/overrides/sales_invoice.py
```diff
@@ -110,67 +117,82 @@ def make_pos_gl_entries(self, gl_entries):
...
-							item=self,
-						)
-					)
+				against_voucher = self.name
+				if self.is_return and self.return_against and not self.update_outstanding_for_self:
+					against_voucher = self.return_against
 
-					# Debit entry to payment mode account
-					payment_mode_account_currency = get_account_currency(payment_mode.account)
+				payment_amount = flt(payment_mode.base_amount) or flt(payment_mode.amount)
```
engahmed1190: **High — this collapses company currency and transaction currency into one variable.**

```python
payment_amount = flt(payment_mode.base_amount) or flt(payment_mode.amount)
```

`base_amount` is company currency; `amount` is transaction currency. `payment_amount` is then written into `credit` / `debit` — fields that must be company currency — as well as into the `*_in_account_currency` fields below. On a foreign-currency POS profile where `base_amount` arrives as 0, a transaction-currency figure lands in a company-currency GL field. That follows from the types alone.

ERPNext gates on `if payment_mode.amount:` and posts `base_amount`; the two are deliberately distinct. Worth restoring that separation.

Worth checking separately (I reasoned about this but did not construct the state): a few lines up, `payment_mode.base_amount -= flt(self.change_amount)`. If that ever lands on exactly 0, the `or` falls through to the un-reduced `amount` and the change gets booked back in — the `or` is what would turn a zero-amount skip into a silent GL imbalance.

POS/src/components/common/AuthorizationDialog.vue
```diff
@@ -0,0 +1,185 @@
...
+		@pointerdown.stop
+	>
+		<FocusScope trapped as-child>
+			<div class="w-full max-w-sm rounded-xl bg-white shadow-xl dark:bg-gray-800">
+				<div class="border-b border-gray-200 px-5 py-4 dark:border-gray-700">
+					<h3 class="text-lg font-semibold text-gray-900 dark:text-gray-100">
+						{{ __("Authorization Required") }}
+					</h3>
+					<p class="mt-1 text-sm text-gray-500 dark:text-gray-400">
+						{{ __("A manager must approve this action.") }}
```
engahmed1190, line 15: **High — the approver is never shown what they are approving.**

This dialog renders "A manager must approve this action.", an approver picker and a PIN box. It never shows the action name, the amount, the customer or the invoice. `state.actionLabel` is captured in `useAuthorization.js` and never rendered anywhere.

The refund amount is supplied by the client (`ReturnInvoiceDialog.vue`, in the `requireAuthorization` context) and becomes the grant's binding ceiling on the server. So a manager who walks to a till and types a PIN has no way to tell whether they are approving a 12.00 refund or a 12,000.00 one — which is the one thing this control exists to prevent.

Render the action, the amount and `return_against` in the dialog body.

pos_next/api/authorization.py
```diff
@@ -0,0 +1,200 @@
...
+
+	return {
+		"rule": rule.name,
+		"ready": sorted(with_pin),
+		"missing_pin": sorted(candidates - with_pin),
+	}
+
+
+@frappe.whitelist()
+@rate_limit(limit=5, seconds=60)
```
engahmed1190: **High — this rate limit keys on IP, so it is one bucket for the whole store.**

`frappe.rate_limiter.rate_limit` defaults to `ip_based=True`, and no `key` is passed, so the identity is `frappe.local.request_ip`. Every till behind one NAT shares a single bucket of 5 approvals per minute. A busy shop hits 429s on legitimate approvals, and the failure mode is a hard block on refunds.

Pass `key="approver"` or set `ip_based=False`. `pin.max_failures` + the lockout is already the brute-force control here, so this decorator does not need to carry that job.

pos_next/api/authorization.py
```diff
@@ -0,0 +1,200 @@
...
+	return {
+		"authorized": True,
+		"grant_token": token,
+		"expires_in": grants.GRANT_TTL,
+		"approved_by": approver,
+	}
+
+
+@frappe.whitelist()
+def set_authorization_pin(user: str, new_pin: str, current_pin: str | None = None) -> dict:
```
engahmed1190, line 170: **Medium — PIN enrolment is self-service and unverified.**

`current_pin` is only required when the user *already has* a PIN. So any authenticated user who sits in an approver Role but has never been issued one can set their own, with no manager involved. The PIN is meant to be a manager-controlled second factor; its enrolment is not controlled at all.

The `else` branch is a second path: `has_permission("User", "write")` means anyone who can edit a User record can overwrite an approver's PIN and then approve as them — and the audit log records the legitimate approver, not them.

Consider `frappe.only_for("System Manager")` for setting another user's PIN, and requiring an explicit grant for first enrolment.

pos_next/authorization/gate.py
```diff
@@ -0,0 +1,113 @@
...
+		approvers = policy.describe_approvers(rule)
+		frappe.throw(
+			_("{0} requires authorization by: {1}").format(frappe._(action.name), approvers)
+			if approvers
+			else _("{0} requires authorization.").format(frappe._(action.name)),
+			title=_("Authorization Required"),
+		)
+
+	if doc is not None:
+		doc.set(FIELD_AUTHORIZED_BY, grant["approver"])
```
engahmed1190, line 116: **Medium — these two fields are never created, so this write silently does nothing.**

`custom_authorized_by` / `custom_authorized_at`: there is no Custom Field fixture, nothing in `patches.txt`, and nothing in `install.py` that creates them. `BaseDocument.set` writes an unknown fieldname straight to `self.__dict__` (`frappe/model/base_document.py:241`) — no error raised, and nothing persisted.

Net effect: the invoice carries no record of who approved it. Only `POS Authorization Log` does.

Ship the two Custom Fields, or delete these two lines — silently writing to a field that does not exist is the worst of the three options.

POS/src/utils/promoApi.js
```diff
@@ -0,0 +1,122 @@
...
+ * Runtime API paths for optional POSNext Promotions.
+ *
+ * pos_next never imports the promotions Python package. Vue may call
+ * posnext_promotions.api.* only when that app's extend_bootinfo flag is present.
+ */
+
+export function isPromotionsAppInstalled() {
+	return Boolean(window.frappe?.boot?.posnext_promotions);
+}
+
```
engahmed1190, line 11: **Medium — this hands authorization to an app that this PR does not ship, while the server still enforces locally.**

`isAuthGateInstalled()` returns true whenever `frappe.boot.posnext_promotions` is set. So on any site with `posnext_promotions` installed, `get_authorizers`, `request_grant` and `get_authorization_policy` are called on `posnext_promotions.api.authorization.*` — while `before_submit` is still enforced by `pos_next.authorization.gate`, reading grants from `pos_next`'s own redis keyspace.

If the satellite's endpoints are absent, or mint tokens in their own namespace, the dialog lists no approvers and `grants.consume` never matches: returns become permanently unsubmittable. Nothing in this repo can verify the other side and no test covers it.

Worth deciding explicitly who owns the gate — move it into the satellite, or make `isAuthGateInstalled()` reflect the gate rather than the promotions app.

pos_next/overrides/sales_invoice.py
```diff
@@ -95,9 +73,38 @@ class CustomSalesInvoice(SalesInvoice):
 	for wallet payment methods marked with is_wallet_payment.
 	"""
 
+	def validate(self):
```
engahmed1190, line 88: **Medium — `_validate_pos_payment_accounts()` is a no-op on first save because of this ordering.**

It runs before `super().validate()`, but `debit_to` is assigned *inside* ERPNext's `validate` — `set_missing_values` / `validate_debit_to_acc` (`sales_invoice.py:711,923`) — and POS never sends `debit_to` (`api/invoices.py:361` says so explicitly).

So the guard `if not self.payments or not self.debit_to: return` short-circuits on every new invoice, and the check only fires on the re-validate at submit. Moving the call after `super().validate()` makes it do what it looks like it does.

pos_next/overrides/sales_invoice.py
```diff
@@ -95,9 +73,38 @@ class CustomSalesInvoice(SalesInvoice):
...
+
+	def on_submit(self):
+		# Re-align after fetch_from so ERPNext's loyalty branch does not run on a
+		# credit note whose original invoice has no loyalty_program.
+		from pos_next.api.sales_invoice_hooks import sync_return_loyalty_program
+
+		sync_return_loyalty_program(self)
+		super().on_submit()
+
+	def _ensure_pos_customer(self):
```
engahmed1190, line 105: **Medium — this silently re-attributes the sale rather than failing.**

`_resolve_pos_customer` falls back to the POS Profile's default customer when `self.customer` does not resolve. If a customer was renamed or deleted, the invoice is quietly reassigned to the walk-in customer and revenue is booked against the wrong party, with nothing in the document saying so.

Throwing when a non-empty `customer` fails to resolve, and falling back only when it was blank to begin with, keeps the convenience without the silent reattribution.

POS/src/stores/posCart.js
```diff
@@ -1050,6 +1131,204 @@ export const usePOSCartStore = defineStore("posCart", () => {
 		return applied;
 	}
 
+	/**
```
engahmed1190: **Medium — satellite behaviour hardcoded in the core cart, in a PR whose premise is the opposite.**

`applyOfflineGiftPool` and `applyOfflineGwpSameItem` are ~200 lines of promotions-app logic living in `pos_next`'s cart, gated at runtime on `isPromotionsAppInstalled()`. This same PR ships `POS/src/utils/offerStrategies.js`, a plugin registry built for exactly this — and only `Accumulative` uses it. The two mechanisms contradict each other.

Three things inside `applyOfflineGiftPool` that also want a second look, since none has a parity test against the server:

- gift distribution is a round-robin over pool SKUs (`poolCodes[i % poolCodes.length]`) that is not derived from any server rule — offline and online can hand the customer different products;
- `paidQty` is computed and then never used, so gifts do not scale with basket size;
- `existingFreeRow.quantity = freeItemsToGive` assigns rather than accumulates.

POS/src/stores/posCart.js
```diff
@@ -426,17 +456,26 @@ export const usePOSCartStore = defineStore("posCart", () => {
 	 * @returns {void}
 	 */
 	function processFreeItems(freeItems) {
-		// Reset free_qty on all non-free items
+		// Restore qty previously carved into same-SKU free rows, then rebuild.
```
engahmed1190, line 459: **Medium — a money invariant held together by four call sites agreeing, with no test.**

`applyOfflineGwpSameItem` carves qty off a paid line onto a same-SKU free line and marks it `gwp_same_item_row: 1`. This loop is the *only* place that restores it. Any path that drops a marked free row without coming through here leaves the customer charged for fewer units than were scanned.

To be clear, I traced the paths and they currently hold — `posDrafts.js` serialises `invoiceItems` verbatim so drafts round-trip the flag, every offer-removal path calls `processFreeItems([])` first, and `InvoiceCart.vue` hides the delete control on free rows via `isLockedFreeRow`. **This is not a live bug.**

But `gwp_same_item_row` is client-only and absent from `buildOfferEvaluationPayload`, so the server can never help reconstruct it, and one new removal path — or someone dropping the `isLockedFreeRow` guard — undercharges silently with nothing to catch it. A unit test on `processFreeItems` carve/restore round-tripping would pin it down, particularly since this logic arguably belongs in the satellite anyway.

pos_next/authorization/gate.py
```diff
@@ -0,0 +1,113 @@
...
+def is_required(action_name: str, pos_profile: str | None = None) -> bool:
+	"""Whether a rule currently gates ``action_name`` on this profile."""
+	return policy.resolve_rule(action_name, pos_profile) is not None
+
+
+def context_from_doc(doc) -> dict:
+	"""Normalised context for a document-triggered action.
+
+	``amount`` is absolute: return invoices carry a negative ``grand_total``, and the
+	binding compares magnitudes so the client and the server agree.
```
engahmed1190, line 78: **Medium — `enforce_context` is unreachable.**

No whitelisted endpoint calls it; only `tests/test_gate.py` does. The cart-level actions it exists for — discount override, price edit — are not wired anywhere, and it is advertised as public API in the package docstring.

Either wire one action to it, or delete it along with the `doctype=None` branch in the registry that exists to serve it.

pos_next/authorization/grants.py
```diff
@@ -0,0 +1,126 @@
...
+def _key(token: str) -> bytes:
+	return frappe.cache().make_key(f"{_KEY_PREFIX}{token}")
+
+
+def _read(key: bytes) -> dict | None:
+	try:
+		raw = frappe.cache().get(key)
+	except Exception:
+		return None
+	return pickle.loads(raw) if raw else None
```
engahmed1190, line 42: **Medium — `pickle` on a cache read.**

The grant is a flat dict of scalars, so `json` does the job. `pickle.loads` on anything reachable from redis is a code-execution primitive if that redis is ever shared or exposed, and there is nothing here that needs pickle's extra reach.

pos_next/authorization/grants.py
```diff
@@ -0,0 +1,126 @@
...
+
+	consumed_by = grant.get("consumed_by")
+	if consumed_by is not None and consumed_by != reference:
+		return None
+
+	if not binding_matches(grant.get("binding") or {}, binding or {}):
+		return None
+
+	grant["consumed_by"] = reference
+	_write(key, grant)
```
engahmed1190, line 109: **Medium — this refreshes the TTL on every consume.**

`_write` sets `ex=GRANT_TTL`, so each successful consume restarts the 180-second window. A grant pinned to a document can be kept alive indefinitely by repeated submits of that document.

If the intent is "the grant lives 180 seconds from issue", the remaining TTL needs preserving here rather than reset.

pos_next/api/authorization.py
```diff
@@ -0,0 +1,200 @@
...
+	return {"success": True, "message": _("PIN updated")}
+
+
+@frappe.whitelist()
+def clear_authorization_pin(user: str) -> dict:
+	"""Remove a user's PIN. System Manager only — this revokes their ability to approve."""
+	frappe.only_for("System Manager")
+
+	pin_store.clear_pin(user)
+	log.record(action="clear_authorization_pin", approver=user, result=log.RESULT_PIN_SET)
```
engahmed1190, line 192: Small, but it is the audit trail: a *removal* is logged as `RESULT_PIN_SET`, so the log records the opposite of what happened.

pos_next/api/customers.py
```diff
@@ -145,17 +148,95 @@ def create_customer(
...
+	]
+	for name in candidates:
+		# Territory allows selecting group nodes in some setups; prefer leaves.
+		if _is_leaf_territory(name):
+			return name
+		# Keep an existing non-empty territory if the leaf check failed only
+		# because the record is missing (avoid wiping a valid leaf passed in).
+		if name and frappe.db.exists("Territory", name) and not _is_leaf_territory(name):
+			continue
+		if name and frappe.db.exists("Territory", name):
```
engahmed1190: This branch is unreachable. Above it:

```python
if _is_leaf_territory(name):                                        # returns
if name and exists(name) and not _is_leaf_territory(name):          # continues
```

By the time control reaches here, `name` either does not exist or is a leaf — and the leaf case already returned. Either the middle `continue` was meant to be something else, or this line can go.

pos_next/integrations/registry.py
```diff
@@ -0,0 +1,67 @@
...
+	for method_path in frappe.get_hooks("pos_next_customer_validators") or []:
+		frappe.get_attr(method_path)(**kwargs)
+
+
+def prepare_customer_doc(customer, **kwargs):
+	sync_external_customer = False
+	for method_path in frappe.get_hooks("pos_next_customer_prepare") or []:
+		result = frappe.get_attr(method_path)(customer, **kwargs)
+		if result:
+			sync_external_customer = True
```
engahmed1190, line 61: These three runners have no `try/except`, unlike every other hook runner in this file. That is right for `validate_customer_create` — a validator should throw.

For `after_customer_insert` it is worth a second look: it runs *after* `customer.insert()`, so a failing external sync now rolls back a customer that was already created. If the external sync is best-effort, it should not be able to undo the local record.

scripts/run_split_smoke_tests.py
```diff
@@ -0,0 +1,36 @@
...
+sys.path.insert(0, os.path.join(BENCH_PATH, "apps"))
+
+
+def bootstrap(site: str):
+	os.chdir(os.path.join(BENCH_PATH, "sites"))
+	frappe.init(site=site)
+	frappe.connect()
+	frappe.set_user("Administrator")
+
+
```
engahmed1190, line 21: The site is hardcoded to `brainwise.dev`, so this script only runs on one machine — it fails immediately on a bench without that site. Worth taking it as an argument or reading `--site`.

Minor, in the same function: `sys.path.insert` after `import frappe` has no effect on the already-imported module.

pos_next/test_split_smoke.py
```diff
@@ -0,0 +1,126 @@
...
+		):
+			result = get_wallet_info("CUST-TEST", "Test Company", pos_profile="POS-TEST")
+		self.assertIn("wallet_enabled", result)
+		self.assertIn("balance_iqd", result)
+		self.assertIn("balance_points", result)
+		self.assertFalse(result["wallet_enabled"])
+
+	def test_prepare_customer_doc_is_generic(self):
+		result = prepare_customer_doc(frappe._dict(name="CUST-TEST"), custom_is_publish=0)
+		self.assertIsInstance(result, bool)
```
engahmed1190: `assertIsInstance(result, bool)` cannot fail — `prepare_customer_doc` returns `sync_external_customer`, which is initialised to `False` and only ever assigned `True`. The test passes whether the hook chain works or not.

Same pattern in `test_bootstrap_settings_hook_extends_flags`, which asserts the presence of the two keys it inserted into `settings` itself a line earlier.

Combined with the `skipTest`s elsewhere in this class, the suite passes on a bench where none of the optional apps are installed while proving very little about the split. `test_no_promotions_import.py` is the model to follow — the AST walk there is a real check.

pos_next/hooks.py
```diff
@@ -157,6 +158,7 @@
 			"pos_next.api.wallet.validate_wallet_payment",
 			"pos_next.overrides.pricing_rule.apply_min_max_price_discounts",
 		],
+		"before_submit": "pos_next.authorization.gate.enforce_document",
```
engahmed1190, line 158: Worth documenting somewhere, because it is a behaviour change with no error message that explains it: this hook is unconditional, and the grant token only ever reaches a document through `api/invoices.py` (`invoice_doc.flags[TOKEN_FLAG]`).

So once a `POS Authorization Rule` exists for returns, **every** other path that submits a POS return — desk UI, `bench execute`, a background job, a bulk tool — is refused, because it carries no flag. That is fail-closed and probably the right default, but the first person to submit a credit note from the desk will have no idea why.

Related: `GRANT_TTL = 180`. Any retry of `submit_invoice` more than three minutes after approval — a slow line, an offline replay via `offline/sync.js` — needs the manager to walk back to the till.

engahmed1190:
## What to finish before re-review

Everything open from [review #5141722425](https://github.com/BrainWise-DEV/POSNext/pull/360#pullrequestreview-5141722425), as a checklist. Each item names the file it lives in and the condition that closes it, so "done" is demonstrable rather than asserted.

**29 items — 13 must close before re-review (Gates 1–2), 5 tests, 8 cleanups, 3 hygiene.**

---

### Gate 1 — Blocking

*The split does not currently do what it claims, and money and approval paths are wrong. These are what make the PR unmergeable rather than merely unpolished.*

- [x] **B1 — Move the optional-app gate from import time to call time** · `pos_next/__init__.py:15,27`
  Delete `_promotions_installed()` and the `if not _promotions_installed():` wrapper. Guard inside each patched function with `optional_apps.promotions_installed()`, as `overrides/pricing_rule.py` and `api/sales_invoice_hooks.py` already do at request time. The duplicate helper must go too — two copies means a future fix lands in one and misses the other.
  **Done when:** the three-site init sequence shows `get_installed_apps` is *not* called during `frappe.init()`, and a site with `posnext_promotions` and one without, served by the same worker, each get the correct patch behaviour. Covered by T1.

- [x] **H2 — Give the wallet-mode cache a TTL and an invalidation hook** · `pos_next/api/wallet.py:182`
  As written, ticking `is_wallet_payment` after the cache warms makes `validate_wallet_payment` return early — the balance check never runs and a customer can overspend their wallet.
  **Done when:** toggling `is_wallet_payment` is reflected on the very next request, with no redis flush.

- [x] **H3 — Stop collapsing company and transaction currency into one variable** · `pos_next/overrides/sales_invoice.py:124`
  Replace `payment_amount = flt(base_amount) or flt(amount)` with ERPNext's separation: gate on `amount`, post `base_amount`. While in there, confirm what happens when the change-amount subtraction above leaves `base_amount` at exactly 0.
  **Done when:** a foreign-currency POS profile posts GL entries whose `credit`/`debit` are company currency and whose `*_in_account_currency` are transaction currency, verified on a real submit. Covered by T2.

- [x] **H4 — Show the approver what they are approving** · `POS/src/components/common/AuthorizationDialog.vue:15`
  Render the action, amount, customer and `return_against` in the dialog body. `state.actionLabel` is already captured in `useAuthorization.js` and thrown away. Without this the manager blind-signs a client-supplied number that becomes the grant's binding ceiling.
  **Done when:** a manager can read the refund amount and source invoice off the dialog without leaving it.

- [x] **H5 — Key the grant rate limit on the approver, not the IP** · `pos_next/api/authorization.py:95`
  Pass `key="approver"` or set `ip_based=False`. Today every till behind one NAT shares a bucket of 5/min, and the failure mode is a hard block on refunds during trade. `pin.max_failures` + lockout already carries the brute-force job.
  **Done when:** five failed approvals for one approver don't stop a different approver on the same network.

- [x] **M1 — Close self-service PIN enrolment** · `pos_next/api/authorization.py:163`
  Two holes: `current_pin` is only required when the user *already has* one, so anyone in an approver Role can self-issue their first PIN; and the `else` branch accepts `has_permission("User", "write")`, so anyone who can edit a User can overwrite an approver's PIN and approve as them — with the log naming the legitimate approver.
  **Done when:** setting another user's PIN requires System Manager, and first enrolment can't be done by the subject alone.

---

### Gate 2 — Correctness, and two open decisions

- [x] **M3 — Decide who owns the authorization gate** 🔀 *decision* · `POS/src/utils/promoApi.js:11`
  `isAuthGateInstalled()` is true whenever `posnext_promotions` is installed, so the frontend calls that app's endpoints while `before_submit` is still enforced by `pos_next`'s gate reading `pos_next`'s redis keyspace. If the satellite's endpoints are missing or namespace their tokens, the dialog lists no approvers, `grants.consume` never matches, and returns become unsubmittable. Nothing in this repo can verify the other half.
  **Decide:** move the gate into the satellite, or make `isAuthGateInstalled()` reflect the gate rather than the promotions app. Record which, and why, in the PR description.

- [x] **M6 — Move Gift Pool and GWP out of the core cart** 🔀 *decision* · `POS/src/stores/posCart.js:1134–1330`
  ~200 lines of satellite behaviour in `pos_next`'s cart, runtime-gated on `isPromotionsAppInstalled()` — in the same PR that ships `offerStrategies.js`, a registry built for exactly this and used only by `Accumulative`. Three bugs come along inside `applyOfflineGiftPool`: the round-robin gift distribution isn't derived from any server rule; `paidQty` is computed and never used so gifts don't scale with basket size; `existingFreeRow.quantity` assigns where it should accumulate.
  **Decide:** if these must stay in core for now, say why the registry doesn't cover them — otherwise the registry is dead weight.

- [x] **M2 — Create the two audit fields, or stop writing to them** · `pos_next/authorization/gate.py:27-28,102-103`
  `custom_authorized_by` / `custom_authorized_at` have no fixture, no patch, nothing in `install.py`. `BaseDocument.set` writes an unknown fieldname to `self.__dict__` and nothing persists, so the invoice records nothing about who approved it.
  **Done when:** either both fields exist after `bench migrate` and a gated return shows the approver on the invoice, or the two `doc.set` lines are gone.

- [x] **M4 — Move `_validate_pos_payment_accounts()` after `super().validate()`** · `pos_next/overrides/sales_invoice.py:76,179`
  `debit_to` is assigned inside ERPNext's `validate` (`sales_invoice.py:711,923`) and POS never sends it, so the `if not self.debit_to: return` guard short-circuits on every new invoice.
  **Done when:** saving a draft whose payment mode points at the receivable account raises immediately, not at submit.

- [x] **M5 — Stop silently re-attributing invoices to the walk-in customer** · `pos_next/overrides/sales_invoice.py:90`
  A renamed or deleted customer falls through to the POS Profile default, booking revenue against the wrong party with nothing saying so.
  **Done when:** an invoice naming a non-existent customer raises rather than submitting under a different one.

- [x] **M8 — Wire `enforce_context` to a real action, or delete it** · `pos_next/authorization/gate.py:64`
  Nothing calls it but `tests/test_gate.py`, yet it's advertised as public API and the registry carries a `doctype=None` branch to serve it.
  **Done when:** either one cart-level action is gated end to end, or the function, its registry branch and its docstring mention are removed.

- [x] **M9 — Swap `pickle` for `json`; stop the grant TTL refreshing itself** · `pos_next/authorization/grants.py:40,44,96`
  The grant is a flat dict of scalars, so pickle buys nothing and costs a code-execution primitive on anything reachable from redis. Separately, `_write` sets `ex=GRANT_TTL` on every consume, restarting the 180s window.
  **Done when:** grants serialise as JSON, and a grant issued 180s ago is dead however many times it was consumed.

---

### Gate 3 — Tests

*`authorization/` is genuinely well tested. Everything else here that touches money is not, and that's exactly where the findings clustered.*

- [ ] **T1 — A test that would have caught B1** · new, alongside `test_no_promotions_import.py`
  Assert the gating is evaluated per call, not per process: importing `pos_next` issues no `get_installed_apps` query, and patch behaviour follows the *current* site rather than the first one initialised.

- [ ] **T2 — Cover `overrides/sales_invoice.py`** · new, `pos_next/overrides/test_sales_invoice.py`
  The largest money-path rewrite in the PR has no tests, and H3, M4 and M5 all live in it. Minimum: GL for a cash sale, a foreign-currency profile and a return; the payment-account guard firing on save; a non-existent customer raising.

- [ ] **T3 — Pin the GWP carve/restore round trip** · new, `POS/src/stores/__tests__/posCart.spec.js`
  Today this invariant holds only because four call sites agree. Test that carving a free row then removing the offer restores the paid quantity exactly, and that a draft save/load preserves it.

- [ ] **T4 — Replace the assertions in `test_split_smoke.py` that cannot fail** · `pos_next/test_split_smoke.py:85` and nearby
  `assertIsInstance(result, bool)` passes whether the hook chain works or not; `test_bootstrap_settings_hook_extends_flags` asserts keys it inserted itself a line earlier. With the `skipTest`s, the suite is green on a bench with none of the optional apps installed while proving almost nothing. `test_no_promotions_import.py` is the model — its AST walk is a real check.

- [ ] **T5 — Prove the offline cart and the server agree on gifts** · new
  The offline round-robin over pool SKUs is a guess at server behaviour. Without parity, an offline sale can hand out a different free product than the same basket would online. Do this wherever the logic ends up after M6.

---

### Gate 4 — Cleanups

- [ ] **L1** — `clear_authorization_pin` logs a removal as `RESULT_PIN_SET`; the audit trail records the opposite of what happened · `api/authorization.py:185`
- [ ] **L2** — Unreachable branch in `_resolve_leaf_territory`: by the time control reaches it, `name` either doesn't exist or is a leaf, and the leaf case already returned · `api/customers.py:229`
- [ ] **L3** — `after_customer_insert` has no `try/except`, so a best-effort external sync can roll back a customer already inserted locally · `integrations/registry.py:61`
- [ ] **L4** — Redundant `sync_return_loyalty_program` in `on_submit`; `validate` already called it before `db_update`, so nothing this changes persists · `overrides/sales_invoice.py:81-87`
- [ ] **L5** — Smoke-test runner hardcodes site `brainwise.dev`; also the `sys.path.insert` after `import frappe` does nothing · `scripts/run_split_smoke_tests.py:20`
- [ ] **L6** — Note the memory ceiling on the full `tabItem Price` snapshot so nobody enables the flag on a large catalogue by surprise · `overrides/test_setup_compat.py`
- [ ] **L7** — `ruff check --fix` clears `I001` in `customers.py:6`, `pricing_rule.py:24`, `sales_invoice.py:10`, plus `RUF100`/`UP038` in `test_split_smoke.py`
- [ ] **L8** — **Confirm the removal of `_combine_packed_qty_for_free_product_bundles` is covered** · `overrides/sales_invoice.py`
  This PR deletes the method that merged duplicate packed rows for free bundle lines. If `patch_packed_item_keying` now covers that case, `test_packed_items_regression.py` should demonstrate it; if it doesn't, the deletion is a regression nobody has noticed yet.

---

### Gate 5 — Before asking for re-review

- [ ] **D1 — Write the PR description.** Six thousand lines reorganising how three apps compose, currently with an empty body. It needs: what the split is and where the boundary now sits; which apps are expected alongside it and which are optional; how the two Gate 2 decisions went; and a manual test plan for a site *with* the satellites and a site *without*.
- [ ] **D2 — Document the fail-closed return behaviour** · `hooks.py:161`
  Once a `POS Authorization Rule` exists, every path that submits a POS return other than `api/invoices.submit_invoice` — desk UI, `bench execute`, background jobs, bulk tools — is refused, because only that path sets the token flag. Probably the right default; documented nowhere, and the error message doesn't explain it. Same note should carry `GRANT_TTL = 180`: a retry after a slow line or an offline replay needs the manager to walk back to the till.
- [ ] **D3 — Rebase on `develop` and re-run everything.** `develop` has moved since merge base `a07f981` (`98b39e5` reworked `get_invoices`, among others). Rebase so the diff shown is the diff that will merge.

---

### Definition of done

Run these and paste the output here. Green, plus Gates 1–2 closed, is what makes this ready for a second look.

```bash
cd POS && yarn test:run              # currently 12 pass — should grow with T1–T5
bench --site <site> run-tests --app pos_next
ruff check pos_next scripts
cd POS && yarn biome lint src

# and the one that proves B1, on a bench with 2+ sites:
python -c "import frappe; frappe.init(site='A'); frappe.destroy(); frappe.init(site='B')"
# -> get_installed_apps must not be called during either init
```

**Deliberately not on this list:** the repo-wide Biome formatting drift (predates this PR), and `get_invoices` — I reported that as a regression and was wrong. It was improved on `develop` after this branch point, and this PR never touched it.

pos_next/__init__.py
MohamedAliSmk: Import-time _promotions_installed() froze the patch gate to whichever
site warmed a multi-tenant worker first, causing silent wrong prices or
double-patching across sites. Always install the patches and defer via
optional_apps.promotions_installed() on each call.

pos_next/overrides/sales_invoice.py
MohamedAliSmk: Stop collapsing base_amount and amount into one payment_amount — credit/debit must stay company currency, and a zero base after change netting must skip instead of falling back to the txn amount.

pos_next/api/wallet.py
MohamedAliSmk: Fixed. The wallet MoP cache no longer lives forever without invalidation.

- Added expires_in_sec=300 as a safety net on pos_next_wallet_payment_modes
- Clear the key from Mode of Payment after_insert / on_update / on_trash

Toggling is_wallet_payment now refreshes the map immediately, so validate_wallet_payment still runs the balance check. TTL covers any missed invalidation path without needing a Redis flush.

POS/src/components/common/AuthorizationDialog.vue
MohamedAliSmk: @MostafaKadry 
please review

pos_next/api/authorization.py
MohamedAliSmk: @MostafaKadry
please review

pos_next/api/authorization.py
MohamedAliSmk: @MostafaKadry
please review

pos_next/authorization/gate.py
MohamedAliSmk: @MostafaKadry
please review

POS/src/utils/promoApi.js
MohamedAliSmk: @MostafaKadry
please review

engahmed1190:
## Cross-reference: this PR and #318

Both are open against `develop` and both touch the same four files, so they need sequencing rather than parallel merges.

**Shared files**

| File | #360 adds | #318 adds |
|---|---|---|
| `POS/components.d.ts` | `AuthorizationDialog` | `ExpenseDialog` |
| `POS/src/pages/POSSale.vue` | menu button + dialog mount | menu button + dialog mount |
| `POS/src/components/sale/InvoiceCart.vue` | wallet/free-item changes | expense button |
| `pos_next/api/bootstrap.py` | `authorization_policy`, `authorization_pin_length` | 2 keys on `result["pos_profile"]` |

`components.d.ts` will conflict outright — each registers a new component. The other three are textually adjacent rather than contradictory, but all four want a human resolving them.

**Suggested order: #318 first, then this one rebases onto it.**

Not a judgement on quality — it is about how far each one's blocker is from done:

- **#318**: the blocker is a config decision — an expense maximum that ships with no default, where `<= 0` is read as "unlimited". Five Gate 1 items, all inside one module.
- **#360** (this PR): the blocker is architectural — the optional-app gate is evaluated once per worker process against an arbitrary site, so the split's central mechanism does not currently engage. That is a rework, and Gate 2 carries two ownership decisions on top (who owns the authorization gate, and whether Gift Pool / GWP belong in the core cart).

Since this PR is the one staying open longer and already needs a rebase, it is the natural place to absorb the four-file conflict.

**When you rebase:** after resolving, confirm both the **Return authorization** and the **POS Expense** entries appear in the POS menu — the two features add sibling buttons to the same menu and the same dialog mount point, so a bad merge silently drops one.

One thing worth knowing before you rebase: `develop` has moved 17 commits past #318's merge base, and commit `98b39e5` reworked `get_invoices` there. I initially flagged that function as a regression in this PR and withdrew it — this PR never touched it, and the rebase will simply take develop's version.

*Reviews and per-PR checklists are in the comments on each. Item counts: 29 here, 21 on #318.*

engahmed1190:
# Review board — PRs #318 and #360

Everything open across both PRs, in one place. **50 items.** Full detail is in each PR's review and checklist comment; this is the consolidated view.

|  | Gate 1 | Gate 2 | Tests | Cleanup + hygiene | **Total** |
|---|---|---|---|---|---|
| **#318** POS expenses | 5 | 7 | 4 | 5 | **21** |
| **#360** Thin develop split | 6 | 7 | 5 | 11 | **29** |

---

## Merge sequence

**First — land #318.** Its blocker is a one-line config decision. Five Gate 1 items, all local to one module. Nothing in it depends on #360.

**Then — rebase #360 onto it.** #360 resolves the four-file collision, since it is the one still needing architectural work. Merging it first would make #318 pay a cost it did not incur.

**Throughout — neither ships without its blocker closed.** Both blockers are unbounded-authority defects, not polish. Everything else on this board can follow.

---

## The two blockers

Everything else here is secondary to these.

### #318 — A blank expense maximum means *unlimited*

Tick one checkbox on a POS Profile and every cashier on it can submit unbounded Journal Entries crediting the till, through a whitelisted endpoint running `ignore_permissions=True`. `posa_maximum_expense_amount` ships with no default, and `<= 0` is read as "no limit".

`custom/pos_profile.json:342` · `api/expenses.py:134`

### #360 — The optional-app gate never engages

Decided once per worker process against whichever site it touched first. Reproduced on a 3-site bench: `fleet.localhost`, which does not have `pos_next` installed, decided `pos_next`'s pricing-patch state for `posnext.local`, which does. The split's central mechanism does not do what it claims.

`pos_next/__init__.py:27`

---

## #318 · POS expenses — 21 items

*Closer to mergeable. Validators are thorough and tested; 17/17 pass. The gaps are all downstream of the money leaving.*

**Gate 1 — Blocking (5)**

- [ ] **B1** Blank maximum must mean "not configured", not unlimited · `custom/pos_profile.json:342`, `api/expenses.py:134`
- [ ] **H1** Lock the shift around check-and-insert · `api/expenses.py:127`
- [ ] **H2** Sum the credit rows, not the writable custom field · `api/expenses.py:154`
- [ ] **H3** Justify and narrow the two `ignore_permissions` bypasses · `api/expenses.py:273, 307`
- [ ] **H4** Add a cancel path, refused once the shift closes · `api/expenses.py:368`

**Gate 2 — Correctness (7)**

- [ ] **M1** Scope the report by company; fix or drop the inert cashier role · `pos_expense_report.py:96`
- [ ] **M2** Link the closing-shift row to its Journal Entry; Data → Link · `pos_closing_shift_expense.json`
- [ ] **M3** Stop resolving the payment account to an arbitrary ledger · `api/expenses.py:249`
- [ ] **M4** 🔀 *decide* — whether non-cash modes may fund an expense · `api/expenses.py:204`
- [ ] **M5** Post overnight expenses to the shift's date · `api/expenses.py:331`
- [ ] **M6** Stop overloading `cheque_no` to carry the shift · `api/expenses.py:334`
- [ ] **M7** Handle multi-currency on the Journal Entry · `api/expenses.py:350`

**Gate 3 — Tests (4)**

- [ ] **T1** One test that builds a real Journal Entry — covers M3, M5, M7 at once
- [ ] **T2** Cover the "no maximum" behaviour that hid B1
- [ ] **T3** Replace the two change-detector tests · `test_expenses.py:159, 213`
- [ ] **T4** Cover the reconciliation seam in `make_closing_shift_from_opening`

**Gate 4–5 — Cleanup & hygiene (5)**

- [ ] **L1** Dead `@expense-created` emit — nothing refreshes · `POSSale.vue:590`
- [ ] **L2** Hardcoded `"USD"` currency default · `ExpenseDialog.vue:146`
- [ ] **L3** Unbounded account list — server-side search instead · `api/expenses.py:307`
- [ ] **D1** Rebase — 17 commits behind develop
- [ ] **D2** Coordinate the four-file collision with #360

---

## #360 · Thin develop split — 29 items

*Further out. The auth gate underneath is well built; the problems are at its edges and in the split mechanism itself.*

**Gate 1 — Blocking (6)**

- [ ] **B1** Move the optional-app gate to call time; delete the duplicate helper · `pos_next/__init__.py:27`
- [ ] **H2** TTL and invalidation on the wallet-mode cache · `api/wallet.py:182`
- [ ] **H3** Stop OR-ing company and transaction currency · `overrides/sales_invoice.py:124`
- [ ] **H4** Show the approver the action, amount and invoice · `AuthorizationDialog.vue:15`
- [ ] **H5** Key the rate limit on approver, not IP · `api/authorization.py:95`
- [ ] **M1** Close self-service PIN enrolment · `api/authorization.py:163`

**Gate 2 — Correctness & decisions (7)**

- [ ] **M3** 🔀 *decide* — who owns the authorization gate · `POS/src/utils/promoApi.js:11`
- [ ] **M6** 🔀 *decide* — move Gift Pool and GWP out of the core cart · `posCart.js:1134–1330`
- [ ] **M2** Create the two audit fields, or stop writing to them · `authorization/gate.py:102`
- [ ] **M4** Move payment-account validation after `super().validate()` · `overrides/sales_invoice.py:76`
- [ ] **M5** Stop silently re-attributing invoices to the walk-in customer · `overrides/sales_invoice.py:90`
- [ ] **M8** Wire `enforce_context` to a real action, or delete it · `authorization/gate.py:64`
- [ ] **M9** `pickle` → `json`; stop the grant TTL refreshing itself · `authorization/grants.py:40, 96`

**Gate 3 — Tests (5)**

- [ ] **T1** A test that would have caught B1
- [ ] **T2** Cover `overrides/sales_invoice.py` — currently zero
- [ ] **T3** Pin the GWP carve/restore round trip
- [ ] **T4** Replace the smoke-test assertions that cannot fail
- [ ] **T5** Prove the offline cart and server agree on gifts

**Gate 4–5 — Cleanup & hygiene (11)**

- [ ] **L1** PIN removal logged as `RESULT_PIN_SET` · `api/authorization.py:185`
- [ ] **L2** Unreachable branch in `_resolve_leaf_territory` · `api/customers.py:229`
- [ ] **L3** Failing external sync rolls back a created customer · `integrations/registry.py:61`
- [ ] **L4** Redundant `sync_return_loyalty_program` in `on_submit`
- [ ] **L5** Smoke runner hardcodes site `brainwise.dev`
- [ ] **L6** Note the memory ceiling on the Item Price snapshot
- [ ] **L7** Clear ruff on the changed files
- [ ] **L8** Confirm the deleted packed-bundle merge is covered
- [ ] **D1** Write the PR description — body is empty
- [ ] **D2** Document the fail-closed return behaviour and the 180s TTL
- [ ] **D3** Rebase on develop

---

## The collision

Both PRs touch `POS/components.d.ts`, `POS/src/components/sale/InvoiceCart.vue`, `POS/src/pages/POSSale.vue` and `pos_next/api/bootstrap.py`. The first will conflict outright — each registers a new component. Whoever merges second resolves all four by hand and re-checks that both features still appear in the POS menu.

## Four decisions, not patches

Marked 🔀 above. Each wants a person, and each belongs in the PR description once settled.

1. **#360** — who owns the authorization gate (`pos_next` or the satellite).
2. **#360** — whether Gift Pool and GWP belong in the core cart, given `offerStrategies.js` exists for exactly that.
3. **#318** — whether non-cash modes may fund an expense.
4. **#318** — what a blank expense maximum should mean.

## What I withdrew

Two findings did not survive checking, and are deliberately **not** on this board:

- **#360 — `get_invoices`.** I reported it as a regression that lost search, date filters and the page-size clamp. Wrong: I diffed `develop..pr360` instead of the merge base. Commit `98b39e5` improved it on `develop` *after* this branch point; the PR never touched it and the rebase will take develop's version.
- **#318 — the report's `POSNext Cashier` grant.** It looked like cross-company exposure. It isn't: `frappe.desk.query_report.run` also checks `has_permission("Journal Entry", "report")` (`frappe/desk/query_report.py:207`), and that role holds no Journal Entry permission on this bench, with none added by the shipped fixtures. A cashier gets refused. The inert role grant is filed as a config bug inside M1; the real half is the missing company scoping.

---

*#318 @ `04034eb` (merge base `9212350`) · #360 @ `5d4c16f` (merge base `a07f981`). Verified on a local bench: Frappe + ERPNext, 3 sites, `dns_multitenant`. #318 backend 17/17 pass; #360 vitest 12/12 pass; ruff and biome run on both.*

pos_next/overrides/sales_invoice.py
MohamedAliSmk: Moved _validate_pos_payment_accounts() to after super().validate().

_ensure_pos_customer() stays before ERPNext’s validate (customer is still required up front). The payment-account check now runs once debit_to exists, so it catches same-account Mode of Payment issues on first save instead of only on submit re-validate.

pos_next/overrides/sales_invoice.py
MohamedAliSmk: Done. `_resolve_pos_customer` now:

- **Blank customer** → still falls back to the POS Profile default (walk-in convenience unchanged)
- **Non-empty customer that doesn’t exist** → throws `Invalid Customer` instead of quietly swapping to the default
- **Neither resolves** → still returns `None`, and `_ensure_pos_customer` throws as before

pos_next/api/authorization.py
MohamedAliSmk: @MostafaKadry
please review

POS/src/stores/posCart.js
MohamedAliSmk: Addressed the architectural Medium.

Gift Pool / GWP offline apply no longer lives in pos_next’s cart behind isPromotionsAppInstalled(). Both register through the same offerStrategies plugin seam as Accumulative:

Host (pos_next): registerProductOfferStrategy / getProductStrategyForOffer; applyOfflineFreeItem only delegates to a registered product strategy, then falls through to the plain free-item path.
Satellite (posnext_promotions): public/pos/offer-strategies.js registers Accumulative (price) plus Gift Pool and GWP (product).
Commits: pos_next 00e17f56, posnext_promotions 5d359e8.

On the three Gift Pool internals — left as-is after checking server parity (allocate_gift_pool_free_items / gift-pool docs):

Round-robin (poolCodes[i % n]) is the server rule. Offline/online skew is stock-aware skip online, not an invented distribution.
paidQty unused for scaling is intentional: free qty is once per cart; paid_qty only gates eligibility.
Assign vs accumulate matches the host free-item re-apply pattern (absolute qty), so re-apply stays idempotent.

POS/src/stores/posCart.js
MohamedAliSmk: Pin same-SKU GWP carve/restore with a unit-tested helper.

Extract syncCartFreeItems so free-row qty restore isn't only enforced by call-site agreement, and add round-trip tests for the silent undercharge case.

pos_next/authorization/gate.py
MohamedAliSmk: @MostafaKadry
please review

pos_next/authorization/grants.py
MohamedAliSmk: @MostafaKadry
please review

pos_next/authorization/grants.py
MohamedAliSmk: @MostafaKadry
please review

pos_next/api/customers.py
MohamedAliSmk: Removed the unreachable block. Behavior is unchanged: leaves return immediately; group or missing names fall through to the next candidate, then any leaf, then "All Territories".

pos_next/integrations/registry.py
MohamedAliSmk: after_customer_insert now logs and continues on hook failure, matching the loyalty/bootstrap runners.

Left alone on purpose:

validate_customer_create — must raise
prepare_customer_doc — runs before insert(), so a failure never leaves an orphaned local Customer; keeping it strict still makes sense unless you want prepare to be best-effort too

scripts/run_split_smoke_tests.py
MohamedAliSmk: Make split smoke tests take --site and fix apps path bootstrap order.

Hardcoding brainwise.dev broke other benches; insert apps on sys.path before importing frappe so the path change actually applies.

pos_next/test_split_smoke.py
MohamedAliSmk: Rewrote test_split_smoke.py so it fails when the seam is broken, not when optional apps are missing.

Replaced tautologies

prepare_customer_doc: AST check that it reads pos_next_customer_prepare, plus mocked hooks proving empty → False, falsy → False, truthy → True (and that kwargs are forwarded).
extend_bootstrap_settings: mocked hook must mutate a fresh {}; empty hooks leave settings unchanged.
Dropped skip-heavy “when installed” checks (doc events, LP API, promotions override) — those belong in the satellite apps. Replaced with a static assert that pos_next/hooks.py does not register Magento handlers.

Kept wallet mocks, no-magento-module import, and bootstrap defaults from DEFAULT_POS_SETTINGS.

pos_next/hooks.py
MohamedAliSmk: @MostafaKadry
please review

MohamedAliSmk:
## Collision with #318 — agreed merge order

**Land #318 first. This PR (#360) rebases second and resolves the four-file collision.**

| File | From #318 (must survive rebase) | From #360 (must keep) |
|---|---|---|
| `POS/components.d.ts` | `ExpenseDialog` | `AuthorizationDialog` |
| `POS/src/pages/POSSale.vue` | POS Expense menu + dialog mount | `:company` on InvoiceCart |
| `POS/src/components/sale/InvoiceCart.vue` | POS Expense empty-cart button | wallet / free-item / promo changes |
| `pos_next/api/bootstrap.py` | expense profile keys | auth policy / pin length / timezone / `extend_bootstrap_settings` |

### After rebase — menu smoke check

- [ ] **POS Expense** in header menu and InvoiceCart empty-cart grid
- [ ] **Return Invoice** still present; gated return still opens `AuthorizationDialog` (`App.vue`)
- [ ] **Promotions** still in ManagementSlider

`components.d.ts`: keep **both** registrations. Do not take only one side of the conflict.

POS/src/components/common/AuthorizationDialog.vue
MostafaKadry: @engahmed1190  I think this is more of a UX improvement than a security/correctness issue in our current flow.

The authorization dialog is not shown to the manager on a separate device or in isolation. The cashier calls the manager over to the cashier's terminal, where the full invoice and its details are already visible. The customer is also physically present with the printed invoice, which is required to process the return, so the manager can verify the printed invoice against the invoice shown on the cashier's screen before approving.

The Auth Required dialog is simply the final approval step after those checks have already been performed.

This flow is also exactly what the business team presented to the customer, and it matches the current customer's agreed requirements. So I would not classify the absence of these details in the authorization dialog as a critical security gap. I agree that showing the action, amount, and return reference could be a useful UX improvement, but I don't think it should be treated as a high-severity issue for this flow.

MostafaKadry:
@MohamedAliSmk @engahmed1190 All the requested issues and comments have been addressed and thoroughly tested. From my side, all three points have been resolved.

@MohamedAliSmk , is there anything else currently blocking the merge of this branch?

engahmed1190:
@MohamedAliSmk @MostafaKadry, please review the action-item checklist in the comments of PR #360 and check off any items that have already been completed. For any remaining items, please add a brief status update.

pos_next/api/authorization.py
MostafaKadry: Gate ownership: Authorization is now fully owned by pos_next; the duplicate posnext_promotions implementation and duplicate before_submit hook were removed.
Core flow: Return Invoice → AuthorizationDialog → PIN validation → submission is verified working end-to-end.
Security: Rate limiting was fixed to use ip:approver instead of IP-only, and PIN lockout now fails closed on cache failures. Brute-force protection remains per-approver via pin.max_failures.
Grant handling: Authorization grants use a 15-minute configurable TTL. The re-stamping-on-retry behavior is intentional and tested.

POS/src/components/common/AuthorizationDialog.vue
MostafaKadry: Approval context: Approvers now see the action, amount, and invoice before entering the PIN. Authorization audit fields were also moved into pos_next.

pos_next/hooks.py
MostafaKadry: The gate was changed to apply only to POS documents and skip framework/non-POS submission contexts, so Desk/back-office/import/framework paths are no longer incorrectly blocked just because they don't carry the grant token

PR #361, opened 2026-09-05
fix(shift-closing): count partial-invoice payments in reconciliation

NotAbdelrahmanelsayed:
## Problem

When a cashier collects a payment against an **existing partially-paid / due Sales Invoice** (e.g. 500 cash on an invoice with an outstanding balance), the amount is recorded correctly as a Payment Entry — but it **does not show up in the Close Shift reconciliation**. The cashier ends up with a cash drawer that looks short by exactly the amount collected on due invoices.

## Root cause

A `reference_no` mismatch between how the payment is created and how shift closing queries it:

| | `reference_no` written |
|---|---|
| Normal POS invoice payment | `<opening shift name>` (e.g. `POSA-OS-24-0000001`) |
| Partial / due invoice payment | `POS-<invoice name>` (fallback) |

`get_payments_entries()` in `pos_closing_shift.py` filters `WHERE reference_no = <opening shift name>`, so partial-invoice payments are never matched.

## Fix

Thread the active opening shift from the POS UI down to the Payment Entry, so it is tagged with the shift name and discovered during reconciliation.

**Backend — `pos_next/api/partial_payments.py`**
- `create_payment_entry()` and `add_payment_to_partial_invoice()` accept an optional `pos_opening_shift`.
- When no explicit `reference_no` is supplied, the shift name is used as `reference_no`. The `POS-<invoice>` fallback is preserved for non-shift contexts.

**Frontend**
- `InvoiceManagement.vue` and `PartialPayments.vue` pass the current shift (`useShift().currentShift?.name`) when submitting a payment.

## Result

Cash (and any mode of payment) collected on due/partial invoices during a shift now appears in the Close Shift reconciliation totals.

## Scope

3 files, +16/−1. Backward compatible — `pos_opening_shift` is optional.

---
_Reopens #300, which was auto-closed by the stale-PR bot for lack of recent activity. No code changes since — recreated to bring it back into review._


engahmed1190:
# Review — PR #361 `fix(shift-closing): count partial-invoice payments in reconciliation`

Branch `feat/partial-payments-shift-reconciliation` @ `cb56e62`, merge base `a10cd40`. 3 files, +16 / −1. It reopens #300, which was closed by the stale bot and never had a human review.

## 1. Decision

**Merge after fixes.** Two changes close it:

1. The server validates the `pos_opening_shift` it is given before writing it into a Payment Entry (B-1).
2. `partial_payments.py` passes `ruff format --check` again, with no new ruff errors (B-2).

The diagnosis is right, and so is the approach. `get_payments_entries()` matches on `reference_no == <opening shift name>`, and partial payments were tagged `POS-<invoice>`, so they never showed up in the closing. The gap is that the shift name comes straight from the browser and nothing checks it. A stale tab tags the payment to a shift that has already closed, and the cash disappears from reconciliation again. That is the same symptom this PR sets out to fix.

## 2. Measured

Run on `pr361`, checked out in the bench app directory. The comparison is against merge base `a10cd40`.

| Gate | base `a10cd40` | pr361 |
|---|---|---|
| `yarn test:run` (vitest) | no test files (the base predates the suite) | same, so nothing to run |
| `yarn lint` (biome) | 205 errors | 205 errors, **0 new**. Both touched `.vue` files already failed format on base |
| `ruff check pos_next` | 40 errors | **42 errors (+2)**: `partial_payments.py:349` UP007, `:709` RUF013 |
| `ruff format --check partial_payments.py` | formatted | **would reformat** (`:708-710`) |
| `bench run-tests --module …test_pos_closing_shift` | — | 0 tests. The module is an empty stub, and no test in the app touches `partial_payments` or `get_payments_entries` |

Checklist sweep over `a10cd40..pr361`:

- **B1:** a new parameter on a whitelisted endpoint, with no validation. This is B-1 below.
- **B2:** the `ignore_permissions` on the Payment Entry predates this PR.
- **B3–B10:** nothing.
- **F1–F7:** nothing. `currentShift` comes from a shared module-level ref, so `.name` on it is correct.

## 3. Blocking findings

### B-1: The client-supplied `pos_opening_shift` is written into a money record without validation

`pos_next/api/partial_payments.py:708` (the new parameter), which becomes `reference_no` at `:454-457`.

The shift closing counts every submitted `Receive` Payment Entry whose `reference_no` equals the shift name (`pos_closing_shift.py:377`). The endpoint accepts whatever string the browser sends:

- **Stale shift, the likely case.** The shift is closed from another tab or device. This tab's `shiftState`, which is persisted in `localStorage` as `pos_shift_data`, still holds the old shift. The cashier collects 500 cash on a due invoice, and the Payment Entry is tagged to a shift whose closing is already done. The 500 appears in no reconciliation. That is the bug this PR fixes, reproduced through a different path.
- **Another cashier's shift.** Any user with write access on Sales Invoice can pass a shift they don't own. That inflates the other cashier's expected cash, and their drawer shows a false shortage.

This is not a regression: invoice submit also trusts a client-supplied shift (`invoices.py:2411`). It still needs fixing here, because this PR is what makes the field count in reconciliation.

**Done when:** `add_payment_to_partial_invoice` checks that the shift exists, has `docstatus = 1` and `status = "Open"`, and has `user = frappe.session.user`. If a shift is given and fails that check, it raises an error that tells the cashier to reload, rather than silently falling back to `POS-<invoice>`. One query does it, and the signature stays the same:

```python
if pos_opening_shift and not frappe.db.exists(
	"POS Opening Shift",
	{"name": pos_opening_shift, "user": frappe.session.user, "status": "Open", "docstatus": 1},
):
	frappe.throw(_("Shift {0} is not open for you. Reload POS and try again.").format(pos_opening_shift))
```

Put this with the other input validation, before the savepoint.

### B-2: `partial_payments.py` now fails `ruff format` and adds 2 ruff errors

`pos_next/api/partial_payments.py:708-710` and `:349`.

The file was formatted on base. The new signature is wrapped over three lines, but it fits on one. `pos_opening_shift: str = None` is an implicit `Optional` (RUF013). The `Optional[str]` at `:349` (UP007) matches the parameters next to it and is fine to keep.

**Done when:** `ruff format --check pos_next/api/partial_payments.py` passes. `ruff check pos_next` reports at most 41 errors, and the RUF013 at `:709` is gone (write the annotation as `str | None = None`).

## 4. Non-blocking

None of these block the merge.

- **There is no test for a money fix.** One test would pin it: `create_payment_entry(..., pos_opening_shift=S)` produces a Payment Entry that `get_payments_entries(S)` returns. This bench can't seed `_Test Company` (it has no outbound network), so I'm not blocking on it. It is still worth adding where fixtures exist.
- **Payments collected with no shift open** still get `POS-<invoice>` and appear in no closing. That is the documented fallback, but the PR body could say so plainly.
- **The explicit `reference_no` branch still wins over the shift.** No frontend path sends `reference_no` today, so this is fine. If a card-reference field is ever added, those payments will drop out of reconciliation again. A comment on the `if reference_no:` branch would warn the next person.

## 5. What's good

- **The root cause is pinned exactly.** The `reference_no` table in the PR body is the whole bug, and I confirmed it against `get_payments_entries`.
- **It is the smallest correct fix.** It reuses the key the closing already matches on, so there's no new field, patch or query in the closing code.
- **There's no double counting.** `_process_invoice` aggregates only the invoice's own `payments` rows, and a Payment Entry never writes to those. So a due invoice created and paid in the same shift is counted once for each real payment.
- **It is backward compatible.** The parameter is optional, the `POS-<invoice>` fallback is kept, and both UI entry points (`InvoiceManagement.vue` and `PartialPayments.vue`) were updated. They are the only two callers.

## 6. Caveats

- The branch base predates the vitest suite, so there was no JS test number to take. I made no browser run. The frontend change is two object keys, and I verified it by reading the code.
- The stale-shift scenario in B-1 comes from reading `useShift.js`, where the state is cleared only in the tab that closed the shift. I did not reproduce it with two live tabs.
- No integration test was run for the Payment Entry path, because `_Test Company` can't be seeded on this bench.

NotAbdelrahmanelsayed:
Updated this PR against current `develop`. The payment helper now rejects a supplied shift unless it is submitted, open, and owned by the cashier, so stale or another cashier's shifts cannot receive due-invoice payments. Added regression tests for valid and invalid shifts and the no-shift fallback. Local checks passed (3 Python tests, 57 frontend tests, production build); GitHub Server and dependency checks are green. Ready for merge.

PR #234, opened 2026-04-07
feat: focus first quick-amount button for keyboard-only payment

NotAbdelrahmanelsayed:
## Summary

- When the payment dialog opens (desktop), the first quick-amount button — always the **exact grand total** — is auto-focused
- Cashier presses **Enter** → amount applied → invoice auto-completed. Zero mouse clicks.
- Mobile custom-amount input: **Enter** applies amount and auto-completes if fully paid
- Numpad **Enter**: auto-completes after adding an amount if fully paid
- Typing **0** + Enter on mobile triggers pay-on-account when credit sale is enabled

## Demo

https://github.com/user-attachments/assets/81180646-4b7d-4d32-aced-c77581056951

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #367, opened 2026-09-05
feat: focus first quick-amount button for keyboard-only payment

NotAbdelrahmanelsayed:
## Summary

- When the payment dialog opens (desktop), the first quick-amount button — always the **exact grand total** — is auto-focused
- Cashier presses **Enter** → amount applied → invoice auto-completed. Zero mouse clicks.
- Mobile custom-amount input: **Enter** applies amount and auto-completes if fully paid
- Numpad **Enter**: auto-completes after adding an amount if fully paid
- Typing **0** + Enter on mobile triggers pay-on-account when credit sale is enabled

## Demo

https://github.com/user-attachments/assets/81180646-4b7d-4d32-aced-c77581056951

---
_Reopens #234, which was auto-closed by the stale-PR bot for lack of recent activity. No code changes since — recreated to bring it back into review._

### DIFF
--- a/POS/src/components/sale/PaymentDialog.vue
+++ b/POS/src/components/sale/PaymentDialog.vue
@@ -1349,7 +1349,23 @@
 						</div>
 						<div class="grid grid-cols-4 gap-1.5">
 							<button
-								v-for="amount in quickAmounts"
+								v-if="quickAmounts.length > 0"
+								ref="firstQuickAmountBtnRef"
+								@click="addCustomPayment(lastSelectedMethod, quickAmounts[0])"
+								@keydown.enter.stop.prevent="handleFirstQuickAmountEnter"
+								:disabled="isQuickAmountDisabled(quickAmounts[0])"
+								:class="[
+									'font-semibold rounded-lg border-2 transition-all',
+									isCompactMode ? 'px-2 py-2 text-sm' : 'px-2 py-2 text-sm',
+									isQuickAmountDisabled(quickAmounts[0])
+										? 'bg-gray-50 border-gray-100 text-gray-300 cursor-not-allowed'
+										: 'bg-white border-gray-200 hover:border-blue-400 hover:bg-blue-50 text-gray-700 hover:text-blue-600'
+								]"
+							>
+								{{ formatCurrency(quickAmounts[0]) }}
+							</button>
+							<button
+								v-for="amount in quickAmounts.slice(1)"
 								:key="amount"
 								@click="addCustomPayment(lastSelectedMethod, amount)"
 								:disabled="isQuickAmountDisabled(amount)"
@@ -1429,6 +1445,7 @@
 										>{{ currencySymbol }}</span
 									>
 									<input
+										ref="mobileAmountInputRef"
 										v-model="mobileCustomAmount"
 										type="number"
 										inputmode="decimal"
@@ -1454,6 +1471,7 @@
 												? 'bg-gray-50 border-gray-100 text-gray-300 cursor-not-allowed'
 												: 'bg-white border-gray-200 focus:ring-1 focus:ring-blue-500',
 										]"
+										@keydown.enter.prevent="handleMobileAmountEnter"
 									/>
 								</div>
 								<button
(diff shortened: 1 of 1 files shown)


MostafaKadry:
## Code Review

### 1. Double-Enter can double a cash payment
**`POS/src/components/sale/PaymentDialog.vue:2260`**

`isQuickAmountDisabled()` never disables the button for cash payment methods, and `handleFirstQuickAmountEnter` has no reentrancy guard. A fast double Enter (key auto-repeat, or Enter pressed again before the dialog unmounts post-`completePayment()`) can call `_upsertPaymentEntry` twice, adding the amount twice. The overpay guard only blocks amounts *greater* than the grand total, not equal, so an exact double-tender can slip through.

### 2. Enter does nothing on non-first quick-amount buttons
**`POS/src/components/sale/PaymentDialog.vue:1368`**

Tab to the 2nd/3rd/4th quick-amount button and press Enter: the global numpad keydown handler (`usePaymentNumpad.js:76-127`) intercepts it, calls `preventDefault()` (killing the native button click), then fires `onEnter()` with the numpad's own value instead of the focused button's amount. The amount is silently never added — this undercuts the PR's "keyboard-only payment" goal for every button except the specially-wired first one.

### 3. Autofocus can silently fail on slow loads
**`POS/src/components/sale/PaymentDialog.vue:3046`**

The focus logic runs once in a `nextTick` when the dialog opens. If `paymentMethods`/`lastSelectedMethod` resolve after that (slow network, first dialog open of the session), the quick-amounts block doesn't exist yet, so `firstQuickAmountBtnRef.value?.focus()` is a no-op with no retry.

### 4. Click vs. Enter behave differently on the same button
**`POS/src/components/sale/PaymentDialog.vue:1354`**

Mouse-clicking the auto-focused button only adds the payment entry; pressing Enter on that same button (`handleFirstQuickAmountEnter`) also auto-completes the sale if fully paid. Same control, different outcome depending on input device.

### 5. Minor: focus logic relies on a hidden-element quirk
**`POS/src/components/sale/PaymentDialog.vue:3043`**

Instead of branching on the already-available `isMobileView`, the code unconditionally calls `focus()/select()` on both the mobile input and desktop button, relying on the fact that a `display:none` element can't take focus.

---

### Suggested manual test steps

1. **Autofocus on open** — Open the payment dialog; verify the first quick-amount button is focused immediately without clicking.
2. **Enter-to-pay on first button** — With that focus, press Enter; verify the payment is added and, if fully paid, the sale auto-completes.
3. **Click vs Enter parity (#4)** — Reopen and click the same button with the mouse; check whether it auto-completes the same way Enter does.
4. **Double-Enter check (#1)** — With Cash selected and a full-payment quick amount, press Enter twice rapidly (or hold to auto-repeat) right as it completes the sale; check for a duplicated/overpaid entry.
5. **Tab to other quick-amount buttons (#2)** — Tab to the 2nd/3rd/4th button and press Enter; confirm the amount is not added.
6. **Slow network (#3)** — Throttle the network so payment methods load slowly, open the dialog immediately, and check whether the first button ends up focused once methods finish loading.
7. **Mobile view** — Switch to mobile layout and verify the custom-amount input (not a quick-amount button) receives focus and its content is selected.

NotAbdelrahmanelsayed:
Thanks for the detailed review @MostafaKadry — all five points are addressed in 08924ad (branch also merged up with `develop`).

**Also found while fixing:** the three Enter handlers this PR originally added read a bare `isSubmitting.value`, but `isSubmitting` is a prop in this component, so it threw a `ReferenceError` exactly on the happy path (payment fully covers the total). The handlers are now replaced by a single `applyPaymentEnter` rule that uses `props.isSubmitting`.

| # | Finding | Fix |
|---|---|---|
| 1 | Double Enter can double a cash payment | Re-entrancy latch in `applyPaymentEnter` (kept up briefly after the sale is emitted) + `event.repeat` ignored for Enter on the buttons, the mobile input and the numpad listener |
| 2 | Enter dead on quick-amount buttons 2–4 | One `v-for` over all quick amounts, `@keydown.enter.stop.prevent` on each, so the window numpad listener no longer swallows it. Digits typed on the numpad still take priority over the focused button |
| 3 | Autofocus fails on slow loads | Retrying post-flush watcher; only marks itself done once focus actually landed |
| 4 | Click vs Enter differ | Intentional: click stays add-only (a mouse slip can't submit an invoice), Enter is the keyboard fast-path. Documented in a template comment |
| 5 | Hidden-element focus quirk | Explicit `isMobileView` branch (mobile input vs first quick-amount button) |

Also added a visible focus ring on the quick-amount buttons — programmatic focus after a mouse click showed no indicator, so a cashier couldn't see which button Enter would fire.

**Behaviour note:** typing `0` + Enter on mobile no longer triggers pay-on-account. With this shared rule, Enter never puts money on a method the cashier didn't choose and never silently turns a sale into customer debt; pay-on-account stays an explicit button action.

**Verified** in a real browser (desktop 1440px + mobile 390px) with the submit endpoint intercepted so nothing was written: autofocus on open; Enter → exactly 1 submit for the exact total, no console errors; two quick Enters and a held key → still 1 submit; Tab→2nd button + Enter → that button's amount (E£150) is applied; 4th button (E£400 on a E£100 sale) correctly raises the existing "Large Overpayment" confirm; typed `150` + Enter → `Cash:150`; mouse click → adds, does not submit; 5s-delayed payment methods → first button focused once they load; mobile → input focused with its value selected, Enter → 1 submit. `vite build` and `biome lint` pass.

PR #374, opened 2026-09-12
Feat/item tax details

tushar-git26:


MostafaKadry:
Heads up @tushar-git26  .while reviewing the `display_item_sku` work, we noticed the `custom_sku` field it depends on isn't shipped anywhere in the app: no `Custom Field` in `pos_next/fixtures/`, and no `item.json` under `pos_next/pos_next/custom/` (unlike `customer.json`, `pos_profile.json`, etc., which do carry this app's other custom fields). On a fresh install, "Display Item SKU" is a no-op until an admin manually adds that field to Item themselves.

Worth flagging in case you meant to include that customization in this PR but it didn't make it in. happy to have you add it if so.

Also worth knowing: POS Next already supports showing an item's own code out of the box, via the existing "Display Item Code" POS Setting. so if a plain item identifier was the goal, that path already exists. That said, we found "Display Item Code" itself has a couple of small pre-existing issues (not from this PR). it wasn't actually wired up to the item list UI, and the backend never sent its value to the frontend at all. We're fixing those in a separate PR, so it'll work as expected shortly.

If you already maintain a `custom_sku` customization elsewhere and it made it into this PR by mistake, please remove it here so we don't end up with duplicate/conflicting field definitions. happy to coordinate on where it should live.

MostafaKadry:
hey @tushar-git26, went through the tax logic in this pr in more detail and found a few things that could make the pos show a different tax split than what actually ends up on the saved invoice.

1. in items.py, _pick_item_tax_row falls back to a blank tax_category row when nothing matches the transaction's tax_category. erpnext itself doesn't do this, it just returns no match. so on a profile with a tax_category set and item tax rows left blank (which is a normal setup), pos will show a resolved template but the saved invoice might end up using the flat header rate instead.

2. attach_item_tax_templates resolves tax_category from the pos profile, but the actual invoice resolves it from the customer. if one profile serves both in state and out of state customers, the on screen split can disagree with the saved invoice.

3. get_item_detail (barcode scan path) never passes tax_category at all, while get_items / get_items_bulk do. so the same item can resolve a different tax template depending on whether it got scanned or picked from the list.

4. in ShiftClosingDialog.vue, gating the eod print behind silent_print also removed the only place that sets eodPrintFailed. so shops without silent print lose the retry button too, not just the auto print.

5. in EditItemDialog.vue, the new backdrop close logic (mousedown/mouseup) can get stuck if the mouseup doesn't land back on the overlay, like dragging out of the browser window. the next unrelated click on the backdrop would then close the dialog and drop whatever was being edited.

6. the item tax row query doesn't select minimum_net_rate / maximum_net_rate, so price banded item tax rows get treated as always valid in the pos preview, even outside their band.

smaller stuff, not blocking:
- get_item_tax_templates loops every enabled item tax template with no caching
- attach_item_tax_templates also queries uncached and runs on every search keystroke
- the item price validity filter is copy pasted in 5 places in items.py, one of them is already missing the orderby
- formatRate in EditItemDialog.vue duplicates formatPercentage in useFormatters.js with different rounding

can you take a look when you get a chance, happy to help if needed.

PR #380, opened 2026-09-17
fix(pos): the till hung off the bottom of the screen under host chrome

engahmed1190:
`POSSale` claimed the whole viewport with an inline `height: 100vh; max-height: 100vh`. Nothing above POS was accounted for, so when a host puts chrome at the top — Frappe's navbar, the Nexus demo countdown banner, anything that pads the body — the page didn't shrink, it **moved**. Its bottom ran off the screen by exactly the height of that chrome, taking the cart's **Checkout** and **Hold** buttons with it.

## Two failure modes, same cause

- **Desktop (48px banner):** the buttons land underneath `POSFooter`, which is `position: fixed` and so doesn't move with the page. They render, they look enabled, and they swallow the click — `elementFromPoint` returns `pos-footer-component`.
- **Phone (86px banner, wraps to two rows):** that's taller than the footer, so the buttons go off the bottom edge entirely — `elementFromPoint` returns `null`.

Either way the cashier can't take a payment, which on a point of sale is the whole job.

Worth noting this is latent without any demo banner: POS pins a 45px footer over the bottom of its own `100vh` page and gets away with it only because the sale layout happens to leave ~45px of slack. The banner just consumes the slack and exposes it.

## The fix

**Size with flex instead of a magic number.** `html`/`body`/`#app` form a column and `POSSale` fills what's left, so the page is whatever the host actually leaves it. 48px of banner, 86px of a two-row one, or nothing at all — all work out on their own, with no constant to keep in sync.

**The footer keeps `position: fixed`, and the integrity checks that enforce it are untouched.** It publishes its measured height as `--pos-footer-h`, and the page reserves that band with `padding-bottom`, so the strip stays pinned and visible without sitting on top of the buttons. The footer owns the number, so restyling it moves the reservation with it rather than stranding a constant elsewhere in the tree.

## Verification

Browser, in a live demo session with the banner up, at three viewports:

| viewport | banner | before | after |
|---|---|---|---|
| 1280×900 | 48px | hit `pos-footer-component`, 48px overflow | Checkout 811–847, clickable, 0 overflow |
| 1440×800 | 48px | hit `pos-footer-component`, 48px overflow | Checkout 711–747, clickable, 0 overflow |
| 390×844 | 86px | 28px below the fold, 86px overflow | Checkout 757–793, clickable, 0 overflow |

Footer stays `position: fixed` and visible in all three. Confirmed end to end with a real Playwright click — which enforces actionability and refuses a covered element — through the payment dialog to a submitted invoice, `ACC-SINV-2026-00015`.

No unit test: the repo has no frontend test files yet, and jsdom does no layout, so it structurally cannot catch this class of bug.

## Follow-up, not in this PR

`POSSale.vue:246` and `:475` still cap inner scroll areas at `calc(100vh - 60px - var(--header-height, 60px))`. Those are `flex-1` children of a `min-h-0` column, so flex governs their size and the cap is now inert — but they're stale `100vh` assumptions that will mislead the next reader. I left them alone rather than change behaviour I hadn't verified.


engahmed1190:
@MostafaKadry Please test this

MostafaKadry:
How to test it?

If your POSSale's screen in pos_next doesn't render the banner, you simulate the host chrome by adding top padding/height to <body> in devtools (this is exactly how the PR's own verification table was produced):

1. Open the POS Sale screen in the browser, open DevTools.
2. In the Elements panel, select <body> and add inline style, e.g.:
padding-top: 48px;   /* simulates a desktop banner */
   or for the two-row mobile case:
padding-top: 86px;
3. Check the cart's Checkout and Hold buttons:
   - Before the fix (checkout main or develop): the buttons run off the bottom of the screen. either hidden under the fixed POSFooter (desktop) or entirely below the fold (mobile), and clicking there either hits pos-footer-component or nothing at all.pos-footer-component or nothing at all.
   - After the fix (this branch, fix/pos-fills-available-viewport): the page shrinks to fit under the simulated banner, POSFooter stays pinned at the bottom via --pos-footer-h, and Checkout/Hold remain visible and clickable.
4. Optionally resize the viewport to a phone size (390×844) with the 86px padding to hit the two-row mobile case called out in the PR.

You can also just toggle the padding-top on/off while on this branch to see the layout reflow live, confirming it's the flex chain (not a fixed 100vh) doing the work.

